PR #24, opened 2019-10-04
solution

Voronok16:
[DEMO](https://voronok16.github.io/react_redux-list-of-todos/)


src/App/App.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import TodoList from '../components/TodoList/TodoList';
+import './App.css';
+
+class App extends React.PureComponent {
```
irenhh: you don't need class component here

src/App/App.js
```diff
@@ -0,0 +1,37 @@
...
+  render() {
+    console.log(this.props.isLoading);
+    if (this.props.isLoading) {
```
irenhh: destructure `this.props`

src/store/store.js
```diff
@@ -0,0 +1,88 @@
...
+const initialState = {
+  data: [],
+  isLoading: false,
+  isReady: true,
+};
+
+export const getData = () => (dispatch) => {
+  store.dispatch(loadData());
+  Promise.all([
+    fetch('https://jsonplaceholder.typicode.com/todos'),
```
irenhh: add url variable

src/store/store.js
```diff
@@ -0,0 +1,91 @@
+import { applyMiddleware, createStore } from 'redux';
+import thunk from 'redux-thunk';
+
+const TODOS_API_URL = 'https://jsonplaceholder.typicode.com/todos';
+const USERS_API_URL = 'https://jsonplaceholder.typicode.com/users';
```
irenhh: one base url is enough and then just add `/todos` or `/users`

PR #31, opened 2019-10-09
solution Khrystyna

khrystynaStyslo:
[Demo](https://khrystynastyslo.github.io/react_redux-list-of-todos/)


src/App.js
```diff
@@ -1,11 +1,66 @@
...
-      <h1>Hello</h1>
-    </div>
-  );
+import { TodoList } from './components/TodoList/TodoList';
+import {
+  store, getTodosWithUsers, loading, loaded,
+} from './store';
+import { getData } from './api/data';
+
+const TODOS_API_URL = 'https://jsonplaceholder.typicode.com/todos';
```
irenhh: create base url and reuse it

src/store/index.js
```diff
@@ -0,0 +1,59 @@
+import { createStore } from 'redux';
+
+// action types
+const ACTION_TYPES = {
+  LOADING: 'LOADING',
```
irenhh: this is not an appropriate name for an action

src/store/index.js
```diff
@@ -0,0 +1,59 @@
...
+  LOADED: 'LOADED',
```
irenhh: the same

src/api/data.js
```diff
@@ -0,0 +1,3 @@
+const URL = 'https://jsonplaceholder.typicode.com/';
```
irenhh: `BASE_URL`

PR #69, opened 2020-03-11
Develop

AlexanderShpilka:
https://alexandershpilka.github.io/react_redux-list-of-todos/
made with redux-thunk

### DIFF
--- a/src/actions/index.ts
+++ b/src/actions/index.ts
@@ -1,0 +1,66 @@
+import { Dispatch } from 'redux';
+
+import {
+  SET_IS_LOADING,
+  SET_DATA_WAS_LOADED,
+  SET_SORT_OPTION,
+  LOAD_DATA,
+  DELETE_TODO,
+} from './types';
+
+import {
+  IsLoadingAction,
+  DataWasLoadedAction,
+  SortOptionAction,
+  SortOption,
+  TodosAction,
+  Todo,
+  User,
+} from '../constants/types';
+
+import {
+  loadTodos,
+  loadUsers,
+} from '../utils/api';
+
+export const setIsLoading = (isLoading: boolean): IsLoadingAction => ({
+  type: SET_IS_LOADING,
+  payload: isLoading,
+});
+
+export const setDataWasLoaded = (): DataWasLoadedAction => ({
+  type: SET_DATA_WAS_LOADED,
+});
+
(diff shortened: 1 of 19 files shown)


src/components/TodoList.css
```diff
@@ -0,0 +1,16 @@
+.todo-item {
+  position: relative;
+  display: flex;
+  flex-direction: column;
+  max-width: 400px;
+  padding: 5px 40px 5px 0;
+  /* padding-top: 5px;
```
irenhh: remove comments

src/components/TodoList.tsx
```diff
@@ -0,0 +1,62 @@
...
+            X
+          </button>
+        </li>
+      ))}
+    </ul>
+  );
+};
+
+const mapStateToProps = (state: AppState) => {
+  return {
```
irenhh: you can write the same without explicit return

src/constants/types.ts
```diff
@@ -0,0 +1,75 @@
...
+  company: Company;
+}
+
+export interface TodoWithUser extends Todo {
+  user: User;
+}
+
+export type SortOption = 'title' | 'completeness' | 'name' | 'default';
+
+export interface IsLoadingAction {
```
irenhh: there is Action type in Redux, use it to create your custom Action type

src/actions/index.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    const todos = initialTodos.map((todo: Todo) => {
+      return {
+        ...todo,
+        user: initialUsers.find(
+          (currentUser: User) => currentUser.id === todo.userId,
+        ) as User,
+      };
+    });
+
```
irenhh: this file is only for actions, move the logic of this function to another place

src/components/TodoList.css
AlexanderShpilka: removed

src/components/TodoList.tsx
AlexanderShpilka: done

src/constants/types.ts
AlexanderShpilka: done

src/actions/index.ts
AlexanderShpilka: done (if I got it right)

PR #98, opened 2020-06-02
solution

DrozhzhynY:
- https://drozhzhyny.github.io/react_redux-list-of-todos/


src/store/index.ts
```diff
@@ -1,51 +1,128 @@
...
-// Selectors - a function receiving Redux state and returning some data from it
-export const isLoading = (state: RootState) => state.loading;
-export const getMessage = (state: RootState) => state.message;
+export const setTodos = (todos: Todo[]) => ({ type: SET_TODOS, todos });
+export const handleError = (errorMessage: string) => ({ type: HANDLE_ERROR, errorMessage });
+export const deleteTodo = (id: number) => ({ type: DELETE_TODO, id });
+export const setStatus = (id: number) => ({ type: SET_STATUS, id });
+export const sortBy = (field: string) => ({ type: SORT_BY, field });
+
+export const isLoading = (state: RootState) => state.isLoading;
```
irenhh, line 32:
```suggestion
export const getIsLoading = (state: RootState) => state.isLoading;
```

src/store/index.ts
```diff
@@ -1,51 +1,128 @@
...
+export const hasError = (state: RootState) => state.errorMessage;
```
irenhh, line 33: this is not a boolean value, this is a function => `getHasError`

the same for others

src/store/index.ts
DrozhzhynY: fixed

PR #108, opened 2020-06-03
solution

Ishchuk-R:
[DEMO LINK](https://Ishchuk-R.github.io/react_redux-list-of-todos/)


Ishchuk-R:
> where are other components?

fixed it

src/App.tsx
```diff
@@ -1,25 +1,54 @@
...
+  const message = useSelector(getMessage) || '';
+  const visibleContent = useSelector(getVisibleContent);
+
+  const loadTodos = async () => {
+    dispatch(startLoading());
+    try {
+      const todosFromServer = await getTodosFromServer();
+
+      dispatch(setTodos(todosFromServer));
+      dispatch(finishLoading(''));
```
irenhh, line 33: what does this method `finishLoading` should do, I can't understand from its name

src/store/index.ts
```diff
@@ -4,24 +4,42 @@ import { composeWithDevTools } from 'redux-devtools-extension';
...
+const VISIBLE_CONTENT = 'VISIBLE_CONTENT';
+const DELETE_TODO = 'DELETE_TODO';
+const SORT_FIELD = 'SORT_FIELD';
 
 // Action creators - a function returning an action object
 export const startLoading = () => ({ type: START_LOADING });
 export const finishLoading = (message = 'No message') => ({ type: FINISH_LOADING, message });
+export const setTodos = (todos: Todo[] = []) => ({ type: TODOS, todos });
+export const setVisibleContent = () => ({ type: VISIBLE_CONTENT });
+export const setDelete = (index: number) => ({ type: DELETE_TODO, index });
```
irenhh, line 17: set delete and delete todo are different things

src/App.tsx
Ishchuk-R: finishLoading  shows that loading are finished and we can see result

Ishchuk-R:
> your sorting buttons sort the array differently every time, it's because you don't save the initial array and rewrite it every time
> 
> pay attention to naming

fixed it

src/store/index.ts
```diff
@@ -6,15 +6,15 @@ const START_LOADING = 'START_LOADING';
...
+const DELETE_TODO_ITEM = 'DELETE_TODO_ITEM';
 const SORT_FIELD = 'SORT_FIELD';
 
 // Action creators - a function returning an action object
 export const startLoading = () => ({ type: START_LOADING });
 export const finishLoading = (message = 'No message') => ({ type: FINISH_LOADING, message });
 export const setTodos = (todos: Todo[] = []) => ({ type: TODOS, todos });
 export const setVisibleContent = () => ({ type: VISIBLE_CONTENT });
-export const setDelete = (index: number) => ({ type: DELETE_TODO, index });
+export const setDeleteItem = (index: number) => ({ type: DELETE_TODO_ITEM, index });
```
irenhh, line 17: just deleteItem is enough

src/App.tsx
irenhh: then name it as setFinishResult because it sets some message and has nothing in common with loading

PR #119, opened 2020-06-11
solution

Vlad-Viddi:
[DEMO LINK](https://vlad-viddi.github.io/react_redux-list-of-todos/)

not finished:
  1. reverse
  2. delete button with functionality


src/App.tsx
```diff
@@ -1,25 +1,20 @@
+/* eslint-disable import/no-duplicates */
```
irenhh, line 1: remove this

src/components/LoadingButtons.tsx
```diff
@@ -0,0 +1,62 @@
+/* eslint-disable import/no-duplicates */
```
irenhh, line 1: the same

src/components/TodosList.tsx
```diff
@@ -0,0 +1,56 @@
...
+        });
+        break;
+
+      default:
+        return [...todos];
+    }
+
+    if (reverse) {
+      sortedTodos.reverse();
+    }
```
irenhh, line 37: this is a bad practice if you sort and then reverse an array, you do iteration twice instead of checking the current direction of sorting and sorting only once

PR #170, opened 2022-02-11
Solution

Valeria-Valuyskaya:
[DEMO LINK](https://Valeria-Valuyskaya.github.io/react_redux-list-of-todos/)


marina-tilniak (commented):
Good job! I have only one comment: when we select the same user again there will be another request to server. You can fix this by adding a check if user id was changed

PR #173, opened 2022-02-11
solution

OleksandrOstrovskyi:
[DEMO LINK](https://OleksandrOstrovskyi.github.io/react_redux-list-of-todos/)


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    const lowerCaseInput = input.toLowerCase();
+
+    return filteredTodos.filter(({ title }) => (
+      title.includes(lowerCaseInput)
+    ));
+  };
+
+  const filteredTodos = filterTodos();
```
Elena-brv, line 67: useMemo

Elena-brv (changes requested):
- user loads every time, even if the same user was already loaded,
add checking
image

PR #256, opened 2022-06-28
redux list

Andriyk0:
[DEMO LINK](https://Andriyk0.github.io/react_redux-list-of-todos/)


vlasiuk-anatolii (approved):
Maybe, should be change style of button "User1".

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <p>No user selected</p>
+    );
+  }
+
+  return (
+    <div className="CurrentUser">
+      <h2 className="CurrentUser__title">
+        <span>
+          Selected user:
+          {user?.id}
```
Anna-Tkachenko, line 22:
```suggestion
          {user.id}
```

src/components/CurrentUser/CurrentUser.tsx
```diff
@@ -0,0 +1,38 @@
...
+        </span>
+      </h2>
+
+      <h3 className="CurrentUser__name">{user?.name}</h3>
+      <p className="CurrentUser__email">{user?.email}</p>
+      <p className="CurrentUser__phone">{user?.phone}</p>
+      <button
```
Anna-Tkachenko, line 29: You've already added check if user is null, so this syntax user? is redundant

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,104 @@
...
+    const userFromServer = await getUserById(id);
+
+    dispatch(setUserAction(userFromServer));
+  };
+
+  const deleteTodoFromServer = async (id: number) => {
+    await deleteTodo(id);
+    const todosFromServer = await getTodos();
+
+    dispatch(setDeleteTodoAction(todosFromServer));
```
Anna-Tkachenko, line 36: will be better to call here add todos action, cause you have the same functionality for both cases

src/store/index.ts
```diff
@@ -1,54 +1,33 @@
...
+      return {
+        ...state,
+        user: action.payload,
       };
 
+    case ActionType.DELETE_TODO:
+      return {
+        ...state,
+        todos: [...action.payload],
+      };
```
Anna-Tkachenko, line 27: this case is redundant
you have this functionality in set todos case

Anna-Tkachenko (approved):

PR #258, opened 2022-06-29
solutionI

EnjSadman:
[DEMO LINK](https://EnjSadman.github.io/react_redux-list-of-todos/)


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+  return (
+    <div className="TodoList">
+      <h2>Todos:</h2>
+      <input
+        type="text"
+        onChange={(event) => {
+          setTimeout(() => {
+            dispatch(setFilterAction(event.target.value));
+          }, 1000);
```
anastasiia-nikita, line 77: 🔥

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const status = useSelector(getStatusSelector);
+  const todos = useSelector(getTodosSelector);
+  const user = useSelector(getUserSelector);
+  const filter = useSelector(getFilterSelector);
+
+  useEffect(() => {
+    const loadTodosFromServer = async () => {
+      let todosFromServer = await getTodos();
+
+      todosFromServer = todosFromServer.filter(el => el.title.includes(filter));
```
anastasiia-nikita, line 32: It is better to bring it to one register to search in Lower & UpperCase

anastasiia-nikita (changes requested):
Cool job!
According to the task add please a delete button: `Additionally add a button next to each item to remove the item.`

Denchepornyuk:
Screenshot 2022-07-04 at 11 50 00

fix styles

src/components/TodoList/TodoList.tsx
Denchepornyuk: agree, you should handle lower and upper cases here

src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+  return (
+    <div className="TodoList">
+      <h2>Todos:</h2>
+      <input
+        type="text"
+        onChange={(event) => {
+          setTimeout(() => {
+            dispatch(setFilterAction(event.target.value));
+          }, 1000);
```
Denchepornyuk, line 79: nope, it's better to implement debounce instead of setTimeout

Denchepornyuk (changes requested):
good job 👍

PR #283, opened 2022-08-10
Develop

AndriiNychaiuk:
- [DEMO LINK](https://AndriiNychaiuk.github.io/react_redux-list-of-todos/)


src/store/todos.ts
```diff
@@ -0,0 +1,28 @@
+import { Todo } from '../types/Todo';
+
+type SetTodosAction = {
+  type: 'todos/SET',
+  payload: Todo[],
+};
+
+export type TodosAction = SetTodosAction;
```
AKomashchuk, line 8: Я думаю це лишній тип, можна залишити лише SetTodosAction

AKomashchuk (changes requested):

AKomashchuk (approved):

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+        case str === Selector.active:
+          return !todo.completed;
+
+        case str === Selector.completed:
+          return todo.completed;
+
+        default:
+          return todo;
+      }
```
jstmpelowycz, line 74: 1) looks weird that you use switch-case this way – you compare two booleans, you'd rather compare `Selector` values only
2) in switch-case, you have to return item of only one type, and here you return either `Todo` or boolean – not OK

src/App.tsx
```diff
@@ -1,28 +1,89 @@
...
+        dispatch(todosActions.setTodos(todosFromServer));
+        setVisibleTodos(todosFromServer);
+      })
+      .finally(() => dispatch(loadingActions.finishLoading()));
+  }, []);
+
+  useEffect(() => {
+    if (selectedTodo) {
+      getUser(selectedTodo.userId)
+        .then(currentUser => setUser(currentUser));
```
jstmpelowycz, line 44:
```suggestion
        .then(setUser);
```

jstmpelowycz (changes requested):

jstmpelowycz (approved):

PR #307, opened 2022-09-05
Develop

fledith:
[DEMO LINK](https://fledith.github.io/react_redux-list-of-todos/)


src/components/TodoList/TodoList.tsx
```diff
@@ -0,0 +1,64 @@
...
+              >
+                {todo.title}
+              </p>
+            </td>
+            <td className="has-text-right is-vcentered">
+              <button
+                data-cy="selectButton"
+                className="button"
+                type="button"
+                onClick={() => (dispatch(actions.setTodo(todo)))}
```
ivan-kadykalo, line 50: Will be better to create handlers before return and call them here.

We will see all the functions that we use in this component and it will be easier to manage them.

src/store/currentTodo.ts
```diff
@@ -0,0 +1,36 @@
...
+  state: null | Todo = null,
+  action: Action,
+): Todo | null => {
+  switch (action.type) {
+    case 'SET_TODO':
+      return action.payload;
+    case 'UNSET_TODO':
+      return null;
+    default:
+      return state;
```
ivan-kadykalo, line 34: You should return a copy of the state with changed keys.
Fix here and in the rest reducers

```suggestion
    case 'SET_TODO':
      return {
      	...state,
      	todos: action.payload,
      };
    case 'UNSET_TODO':
      return {
      	...state,
      	todos: null,
      };
    default:
      return state;
```

ivan-kadykalo (changes requested):

src/store/currentTodo.ts
fledith: in my case, I have no state with key todos, my state is todos and in this case it is unnecessary in my opinion

src/components/TodoList/TodoList.tsx
fledith: I cant bcz my todo is only in todos.map and if I create handleFunction it is just adding a few new lines, and anyway  I need to write function in onClick

PR #349, opened 2022-10-24
react_redux-list-of-todos

MaxManis:
- [DEMO LINK](https://MaxManis.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,6 +1,22 @@
 import React from 'react';
+import { useDispatch } from 'react-redux';
+import { actions as filterActions } from '../../features/filter';
 
 export const TodoFilter: React.FC = () => {
+  const dispatch = useDispatch();
+
+  const onSetStatus = (newStatus: string) => (
+    dispatch(filterActions.setStatus(newStatus))
+  );
+
+  const onSetQwery = (newQwery: string) => (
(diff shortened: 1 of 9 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,145 @@
...
+    setIsLoading(true);
+
+    getTodos()
+      .then((response) => {
+        dispatch(todosActions.setTodos(response));
+      })
+      .catch(() => {
+        throw new Error('Error');
+      })
+      .finally(() => setIsLoading(false));
```
Denchepornyuk, line 62: async await

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,14 +1,37 @@
...
+
+  const dispatch = useDispatch();
+
+  const onSetStatus = (newStatus: string) => (
+    dispatch(filterActions.setStatus(newStatus))
+  );
+
+  const onSetQwery = (newQwery: string) => (
+    dispatch(filterActions.setQwery(newQwery))
+  );
```
Denchepornyuk, line 18: it's better to not store in redux variables which you use only in one component, you can store them in state

Denchepornyuk (approved):
Screenshot 2022-10-27 at 15 34 46

usually we create files like actions and selectors when we work with redux
It will be easier to understand where you store action, selectors and reducers. Reducer you can store in index.ts file

PR #352, opened 2022-10-31
Done

Oleksii-Blyzniuk:
https://Oleksii-Blyzniuk.github.io/react_redux-list-of-todos/

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,208 +1,107 @@
-/* eslint-disable max-len */
-import React from 'react';
+import classNames from 'classnames';
+import React, { useMemo } from 'react';
+import { useDispatch } from 'react-redux';
+import { useAppSelector } from '../../app/hooks';
+import { actions as currentTodoActions } from '../../features/currentTodo';
+import { FilterStatus } from '../../features/filter';
+import { Todo } from '../../types/Todo';
 
 export const TodoList: React.FC = () => {
+  const todos = useAppSelector(state => state.todos);
+  const filter = useAppSelector(state => state.filter);
+  const dispatch = useDispatch();
+
+  const getVisibleTodos = () => {
+    let filteredTodos: Todo[] = [];
+
+    switch (filter.status) {
+      case FilterStatus.ALL:
+        filteredTodos = todos;
+        break;
+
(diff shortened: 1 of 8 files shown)


polosanya (changes requested):
Filter feature doesn't work
image
Before making a Pull Request (or a new commit to an already existing PR), you need to write the `npm test` command in the terminal and make sure that all tests pass. After that, the mentor will be able to review the homework. If you have questions please ask them in fe_chat

Anastasiia-Svintsova (changes requested):
There is something strange going on with the filter. Doesn't filter everywhere including the search string. After the value in the search string is cleared, the old value is still applied. Check the video below and compare it with the working example
Animation

Bohdan-Maksymiuk (approved):

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,86 @@
...
+  const todos = useSelector(TODOS_SELECTORS.getPreparedTodos);
+  const selectedTodo = useSelector(TODO_SELECTORS.getCurrentTodo);
 
-            <td className="is-vcentered">
-              <span className="icon" data-cy="iconCompleted">
-                <i className="fas fa-check" />
-              </span>
-            </td>
+  const selectedTodoId = selectedTodo?.id;
+  const isLoading = false;
```
witflash, line 15: Why do you need this constant?

src/features/todos.ts
```diff
@@ -1,9 +1,61 @@
...
+  getTodos: (state: RootState) => state.todos,
+  getPreparedTodos: (state: RootState) => {
+    const { todos, filter } = state;
+    const { query, status } = filter;
+
+    let preparedTodos: Todo[];
+
+    switch (status) {
+      case 'completed':
+        preparedTodos = [...todos].filter(todo => todo.completed);
```
witflash, line 30: You don't need to create a new array with spread operator because `filter` method always returns a new array.

sTorba24 (approved):

witflash (approved):
👍🏼

qoa11a (approved):

agniya-i (approved):

polosanya (approved):

polosanya (approved):

PR #354, opened 2022-11-12
 react_redux-list-of-todos

ievgen-volkov:
https://ievgen-volkov.github.io/react_redux-list-of-todos/


src/features/filter.ts
```diff
@@ -1,10 +1,47 @@
...
+const query = (value: string): QueryAction => ({
+  type: 'filter/Query',
+  payload: value,
+});
+
+export const actions = { status, query };
+
+type State = {
+  status: Status,
+  query: string,
```
SanyaBratashchuk, line 21: split actions, reducers, selectors to the different files

src/features/todos.ts
```diff
@@ -1,9 +1,24 @@
...
+const todosAdd = (todos: Todo[]): Actions => ({
+  type: 'todos/ADD',
+  payload: todos,
+});
+
+export const actions = { todosAdd };
 
-const todosReducer = (): Todo[] => {
-  return [];
+const todosReducer = (todos: Todo[] = [], action: Actions): Todo[] => {
```
SanyaBratashchuk, line 15:
```suggestion
const todosReducer = (todos: Todo[] = [], action: Actions): Todo[] => {
```
Reducer should take state as a Param State should be an object

SanyaBratashchuk: Also you can hold isLoading or/and error state

src/App.tsx
```diff
@@ -1,14 +1,29 @@
...
+  const dispatch = useAppDispatch();
+  const currentTodo = useAppSelector(state => state.currentTodo);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+
+  useEffect(() => {
+    setIsLoading(true);
+
+    getTodos()
+      .then(todos => dispatch(todoActions.todosAdd(todos)))
+      .finally(() => setIsLoading(false));
```
SanyaBratashchuk, line 24: You should use async actions

SanyaBratashchuk (changes requested):
Add all files

src/features/currentTodo.ts
```diff
@@ -28,7 +28,10 @@ const currentTodoReducer = (
   action: Action,
 ): State => {
   switch (action.type) {
-    // Implement all actions here
+    case 'currentTodo/SET':
```
polosanya, line 31: Would be great not to hardcode this value too

src/features/filter/filter.ts
```diff
@@ -0,0 +1,26 @@
...
+  status: 'all',
+  query: '',
+};
+
+const filterReducer = (
+  filters: State = initialState,
+  action: Action,
+): State => {
+  switch (action.type) {
+    case 'filter/Status':
```
polosanya, line 13: Same here

polosanya (changes requested):
Almost perfect! Let's improve your code a bit

src/components/TodoList/TodoList.tsx
```diff
@@ -1,13 +1,52 @@
...
+  const status = useAppSelector(state => state.filter.status);
+  const query = useAppSelector(state => state.filter.query);
+
+  const visibleTodos = useMemo(() => {
+    let newTodos = todos.filter(todo => {
+      switch (status) {
+        case Status.ACTIVE:
+          return !todo.completed;
+        case Status.COMPLETED:
+          return todo.completed;
```
loralevitska, line 21:
```suggestion
          return todo.completed;
          
```

PR #378, opened 2022-12-24
first commit

Gintako666:
[DEMO LINK](https://gintako666.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/features/filter.ts
+++ b/src/features/filter.ts
@@ -1,10 +1,59 @@
-export const actions = { /* put action creators here */};
-
-const filterReducer = () => {
-  return {
-    query: '',
-    status: 'all',
-  };
+// export type Status = 'all' | 'active' | 'completed';
+type Filters = {
+  query: string,
+  status: string,
 };
+
+type StatusAction = {
+  type: 'filters/Status'
+  query: string,
+  status: string
+};
+
+type QueryAction = {
+  type: 'filters/Query'
+  query: string,
+  status: string
+};
+
+const editQuery = (filters: Filters, value: string): QueryAction => ({
+  ...filters,
+  type: 'filters/Query',
+  query: value,
+});
+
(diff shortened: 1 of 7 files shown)


src/features/todos.ts
```diff
@@ -1,9 +1,25 @@
...
+type SetTodos = { type: 'todos/Set', todos: Todo[] };
+type ErrorSetTodos = { type: 'todos/Error' };
+type Actions = SetTodos | ErrorSetTodos;
 
-const todosReducer = (): Todo[] => {
-  return [];
+const setTodos = (todos: Todo[]): SetTodos => ({ type: 'todos/Set', todos });
+const errorSetTodos = (): ErrorSetTodos => ({ type: 'todos/Error' });
+
+const todosReducer = (todos: Todo[] = [], action: Actions): Todo[] => {
```
SanyaBratashchuk, line 10: state should be an object

src/features/todos.ts
```diff
@@ -1,9 +1,25 @@
...
+  switch (action.type) {
+    case 'todos/Set':
+      return action.todos;
+
+    case 'todos/Error':
```
SanyaBratashchuk, line 15: create an Enum for actions

src/features/filter.ts
```diff
@@ -1,10 +1,59 @@
...
+});
+
+type Actions = StatusAction | QueryAction;
+
+const startSilters: Filters = {
+  query: '',
+  status: 'all',
+};
+
+const filterReducer = (filters: Filters = startSilters, actions: Actions) => {
```
SanyaBratashchuk, line 38: In my opinion, you don't need to create a separate store for filters, it can be part of todosStore, because if more entities appear in the app that will also have filtering, it can be a problem

SanyaBratashchuk (changes requested):

PR #396, opened 2023-01-21
Implement with redux

UmizDemud:
[DEMO LINK](https://UmizDemud.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,210 +1,103 @@
-/* eslint-disable max-len */
-import React from 'react';
+import classNames from 'classnames';
+import { useCallback } from 'react';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { Todo } from '../../types/Todo';
 
-export const TodoList: React.FC = () => {
+export const TodoList = () => {
+  const todos = useAppSelector(state => state.todos);
+  const currentTodo = useAppSelector(state => state.currentTodo);
+  const filter = useAppSelector(state => state.filter);
+  const dispatch = useAppDispatch();
+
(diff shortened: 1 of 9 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,103 @@
...
+            .filter(todo => !todo.completed
+                && todo.title.toLowerCase().includes(cleanQuery));
+        case 'completed':
+          return (todos)
+            .filter(todo => todo.completed
+                && todo.title.toLowerCase().includes(cleanQuery));
+        default:
+          return (todos)
+            .filter(todo => todo.title.toLowerCase().includes(cleanQuery));
+      }
```
meforoma, line 32: could be quite shorter:
please check whether this would work @UmizDemud 
```suggestion
      return todo.filter(({ title, completed }) => (
        title.toLowerCase().includes(cleanQuery)
        && switch (filter?.status) {
          case 'active':
            return !completed;
          case 'completed':
            return completed;
          default:
            return true;
        }
      ))
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,103 @@
...
-            <td className="is-vcentered">9</td>
-            <td className="is-vcentered"> </td>
-
             <td className="is-vcentered is-expanded">
-              <p className="has-text-danger">molestiae perspiciatis ipsa</p>
+              <p className={classNames(
+                { 'has-text-success': todo.completed },
+                { 'has-text-danger': !todo.completed },
+              )}
+              >
```
meforoma, line 77: no more concerns for the `classNames` package? 🙂

meforoma (approved):
good one @UmizDemud 
check the few comments i wrote.

PR #404, opened 2023-01-29
Done

ruslik87:
[DEMO LINK](https://ruslik87.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/ButtonModal/ButtonModal.tsx
+++ b/src/components/ButtonModal/ButtonModal.tsx
@@ -1,0 +1,17 @@
+import React from 'react';
+import { useAppDispatch } from '../../app/hooks';
+import { actions as currTodoActions } from '../../features/currentTodo';
+
+export const ButtonModal: React.FC = () => {
+  const dispatch = useAppDispatch();
+
+  return (
+    // eslint-disable-next-line jsx-a11y/control-has-associated-label
+    <button
+      type="button"
+      className="delete"
+      data-cy="modal-close"
+      onClick={() => dispatch(currTodoActions.removeTodo())}
+    />
+  );
+};
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,6 +1,24 @@
 import React from 'react';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { actions as filterActions } from '../../features/filter';
+import { Status } from '../../types/Status';
+
+const statuses = [Status.ALL, Status.ACTIVE, Status.COMPLETED];
 
 export const TodoFilter: React.FC = () => {
+  const dispatch = useAppDispatch();
+  const { status, query } = useAppSelector(state => state.filter);
+
+  const handleChangeStatus = (e: React.ChangeEvent<HTMLSelectElement>) => (
+    dispatch(filterActions.setStatusAction(e.target.value as Status))
+  );
+
+  const handleChangeQuery = (e: React.ChangeEvent<HTMLInputElement>) => (
+    dispatch(filterActions.setQueryAction(e.target.value.trimStart()))
+  );
+
+  const handleClearQuery = () => dispatch(filterActions.setQueryAction(''));
+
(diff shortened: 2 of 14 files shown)


src/components/ButtonModal/ButtonModal.tsx
```diff
@@ -0,0 +1,17 @@
...
+export const ButtonModal: React.FC = () => {
+  const dispatch = useAppDispatch();
+
+  return (
+    // eslint-disable-next-line jsx-a11y/control-has-associated-label
+    <button
+      type="button"
+      className="delete"
+      data-cy="modal-close"
+      onClick={() => dispatch(currTodoActions.removeTodo())}
```
Volodymyr-Melnyk-Fron10der, line 14: I think it is too long/complicated method for writing it inline, move it into certain function `handleDeleteTodo `for example

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,17 +1,41 @@
 import React from 'react';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { actions as filterActions } from '../../features/filter';
+import { Status } from '../../types/Status';
+
+const statuses = [Status.ALL, Status.ACTIVE, Status.COMPLETED];
```
Volodymyr-Melnyk-Fron10der, line 6: 1. constants are named with uppercase
2. constants should be placed in separated file, for example `src -> constants -> statuses.ts` or even in your types/status.ts

```suggestion
const Statuses = [Status.ALL, Status.ACTIVE, Status.COMPLETED];
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,106 @@
...
+          return completed;
 
-            <td className="is-vcentered is-expanded">
-              <p className="has-text-success">illo est ratione doloremque quia maiores aut</p>
-            </td>
+        default:
+          return true;
+      }
+    });
+  });
```
Volodymyr-Melnyk-Fron10der, line 30: Try to keep `single responsibility pattern` in your code, So useSelector is only have to SELECT data, that's all. All other operations should be executed outside selector

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,106 @@
...
-                  <i className="far fa-eye" />
-                </span>
-              </button>
-            </td>
-          </tr>
-        </tbody>
-      </table>
+  return (
+    <>
+      {sortedTodos.length === 0
```
Volodymyr-Melnyk-Fron10der, line 34:
```suggestion
      {!sortedTodos.length
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,106 @@
...
+                      onClick={() => (
+                        dispatch(currTodoActions.setTodo(todo))
+                      )}
+                    >
+                      <span className="icon">
+                        <i className={classNames(
+                          'far',
+                          {
+                            'fa-eye': selectedTodo?.id !== todo.id,
+                            'fa-eye-slash': selectedTodo?.id === todo.id,
```
Volodymyr-Melnyk-Fron10der, line 92: You can move this condition into variable to not repeat yourself
```suggestion
                            'fa-eye': !isCurrentTodo,
                            'fa-eye-slash': isCurrentTodo,
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,89 @@
...
 
-      <Loader />
+      {user && (
+        <div className="modal-card">
+          <header className="modal-card-head">
+            <div
+              className="modal-card-title has-text-weight-medium"
+              data-cy="modal-header"
+            >
+              {`Todo #${id}`}
```
Volodymyr-Melnyk-Fron10der, line 61:
```suggestion
              Todo #{id}
```

Volodymyr-Melnyk-Fron10der (changes requested):
Great! But there are some possible fixes

PR #409, opened 2023-02-02
Redux list of TODOs

mr-valentos:
[DEMO LINK](https://mr-valentos.github.io/react_redux-list-of-todos/)


src/App.tsx
```diff
@@ -1,14 +1,40 @@
...
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { getVisibleTodos } from './helper';
+import { useAppDispatch, useAppSelector } from './app/hooks';
+import { actions as todosActions } from './features/todos';
 
 export const App: React.FC = () => {
+  const dispatch = useAppDispatch();
+  const todosRedux = useAppSelector(state => state.todos);
```
romanblyashuk94, line 20:
```suggestion
  const todos = useAppSelector(state => state.todos);
```

src/App.tsx
```diff
@@ -1,14 +1,40 @@
...
+  const currentTodo = useAppSelector(state => state.currentTodo);
+  const { query, status } = useAppSelector(state => state.filter);
```
romanblyashuk94, line 22: you dont use this state in App component (You are just passing them to Filter component)

so, it will be better to get them inside Filter component

src/App.tsx
```diff
@@ -17,18 +43,30 @@ export const App: React.FC = () => {
...
-              <Loader />
-              <TodoList />
+              {todosRedux.length ? (
+                <TodoList
+                  todos={visibleTodos}
+                  selectedTodoId={currentTodo?.id || 0}
+                />
+              ) : (
+                <Loader />
+              )}
```
romanblyashuk94, line 60: if server will return you empty array - loader always will be visible.

use isLoading state

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,82 @@
...
+
+interface Props {
+  todos: Todo[];
+  selectedTodoId: number;
+}
+
+export const TodoList: React.FC<Props> = memo(({
+  todos,
+  selectedTodoId,
+}) => {
```
romanblyashuk94, line 14: again. You dont need to use props passing

redux state is avalible from any component

just take todos and currentTodo from global state

src/features/filter.ts
```diff
@@ -1,10 +1,47 @@
-export const actions = { /* put action creators here */};
+type QueryFilterAction = {
+  type: 'queryFilter/SET';
+  payload: string;
+};
+
+const setQueryFilter = (query: string): QueryFilterAction => ({
+  type: 'queryFilter/SET',
```
romanblyashuk94, line 7: its better to use prefix for action types names same as reduser name or part of state name
fix in all cases
```suggestion
  type: 'filter/SET_QUERY',
```

romanblyashuk94 (changes requested):

PR #412, opened 2023-02-07
Develop

artem-hirzhev:
[DEMO LINK](https://artem-hirzhev.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,14 +1,28 @@
 import React from 'react';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { actions as filterActions } from '../../features/filter';
+import { Status } from '../../types/Status';
 
 export const TodoFilter: React.FC = () => {
+  const filter = useAppSelector(state => state.filter);
+  const dispatch = useAppDispatch();
+
   return (
     <form
       className="field has-addons"
-      onSubmit={event => event.preventDefault()}
+      onSubmit={e => e.preventDefault()}
     >
       <p className="control">
         <span className="select">
-          <select data-cy="statusSelect">
+          <select
+            data-cy="statusSelect"
+            value={filter.status}
+            onChange={
+              (e) => dispatch(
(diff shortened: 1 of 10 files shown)


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,14 +1,28 @@
...
+            value={filter.status}
+            onChange={
+              (e) => dispatch(
+                filterActions.setStatus(e.target.value as Status),
+              )
+            }
+          >
             <option value="all">All</option>
             <option value="active">Active</option>
             <option value="completed">Completed</option>
```
ivbutsykin, line 28: Will be nice move filter types in constants and reuse it everywhere

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -21,19 +35,24 @@ export const TodoFilter: React.FC = () => {
           data-cy="searchInput"
           type="text"
           className="input"
+          value={filter.quary}
```
ivbutsykin, line 38: Typo, should be `query`)

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,69 @@
...
 
-      <div className="modal-card">
-        <header className="modal-card-head">
-          <div
-            className="modal-card-title has-text-weight-medium"
-            data-cy="modal-header"
-          >
-            Todo #3
-          </div>
+      {(!isLoading && user) && (
```
ivbutsykin, line 32: I recommend add check on `user` on with `isLoading`. You can cover one more case in such way (when you don't have user and loading complete):
```
{
  isLoading || !user ? <Loader /> : ...
}
````

ivbutsykin (approved):
Nice work!

tetiana-ivanchenko (changes requested):
fix prevs comments

Volodymyr-Melnyk-Fron10der (approved):
Good👍 Check my comments to keep it in mind but you shouldn't request review anymore

PR #416, opened 2023-02-08
solution

lefteris-L:
[DEMO LINK](https://lefteris-l.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,210 +1,108 @@
-/* eslint-disable max-len */
-import React from 'react';
+import cn from 'classnames';
+import { actions } from '../../features/currentTodo';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { Todo } from '../../types/Todo';
 
 export const TodoList: React.FC = () => {
+  const { todos, currentTodo, filter } = useAppSelector(state => state);
+  const { query, status } = filter;
+  const { setTodo, removeTodo } = actions;
+  const dispatch = useAppDispatch();
+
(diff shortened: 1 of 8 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,108 @@
...
+    const hasQuery = RegExp(query.trim(), 'i').test(title);
+    let hasStatus = true;
+
+    if (status === 'active') {
+      hasStatus = !completed;
+    }
+
+    if (status === 'completed') {
+      hasStatus = completed;
+    }
```
naziamoff, line 22: enum works better for the string values, easier to maintain and less likely to have a typo

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,108 @@
...
+
+    return hasQuery && hasStatus;
+  });
+
+  const toggleCurrentTodo = (todo: Todo) => () => {
+    if (currentTodo?.id === todo.id) {
+      dispatch(removeTodo());
+    } else {
+      dispatch(setTodo(todo));
+    }
```
naziamoff, line 32: check whether it works this way, but it seems more ergonomic this way to me
```suggestion
  dispatch(
    currentTodo?.id === todo.id
      ? removeTodo()
      : setTodo(todo)
  )
```

naziamoff (approved):

src/components/TodoList/TodoList.tsx
lefteris-L: I think it works and yes this is better, thanks

PR #417, opened 2023-02-09
solution

AnnyKoval:
[DEMO LINK](https://AnnyKoval.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,210 +1,122 @@
-/* eslint-disable max-len */
 import React from 'react';
+import classNames from 'classnames';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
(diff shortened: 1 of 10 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,122 @@
...
+                className="button"
+                type="button"
+                onClick={() => dispatch(
+                  currentTodoActions.setTodo({
+                    id,
+                    title,
+                    completed,
+                    userId,
+                  }),
+                )}
```
VitaliyBondarenko1982, line 105: Anonimous function in JSX.
Better to make function for it and pass just reference to onClick prop

VitaliyBondarenko1982 (approved):
Nice job!

Make just little comment about anonimous function, but it could be approved

PR #428, opened 2023-02-15
solution

Andrii-Kuzmenko:
[DEMO LINK](https://Andrii-Kuzmenko.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/features/filter.ts
+++ b/src/features/filter.ts
@@ -1,10 +1,64 @@
-export const actions = { /* put action creators here */};
+type Filter = {
+  query: string,
+  status: string,
+};
 
-const filterReducer = () => {
-  return {
-    query: '',
-    status: 'all',
-  };
+type FilterStatus = {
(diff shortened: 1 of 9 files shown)


src/features/filter.ts
```diff
@@ -1,10 +1,64 @@
...
+  status: string,
+};
+
+type Action = FilterStatus
+| FilterQuery
+| FilterClear;
+
+const filterStatus = (status: string, query = ''): FilterStatus => ({
+  type: 'filter/STATUS',
+  query,
```
slavanosatskyi, line 28: as a common practice action should contain two fields - type and payload. The same as you did for setTodos action

slavanosatskyi (changes requested):
Nice work, I see only one minor comment to be fixed

sTorba24 (approved):
Good work!

PR #459, opened 2023-03-30
Redux list

Mykola-Hadupiak:
[REDUX LIST](https://Mykola-Hadupiak.github.io/react_redux-list-of-todos/)


src/App.tsx
```diff
@@ -1,14 +1,30 @@
...
+  const [isLoading, setIsLoading] = useState(true);
+  const selectedTodo = useAppSelector(state => state.currentTodo);
+  const dispatch = useAppDispatch();
+
+  useEffect(() => {
+    (async () => {
+      const todos = await getTodos();
+
+      dispatch(TodosDispatch.addTodos(todos));
+      setIsLoading(false);
```
Volodymyr-Melnyk-Fron10der, line 24: what if request was failed? Add try catch block and set loading state in the finally block

src/api.ts
```diff
@@ -16,7 +16,13 @@ function get<T>(url: string): Promise<T> {
 
   return wait(300)
     .then(() => fetch(fullURL))
-    .then(res => res.json());
+    .then(response => {
+      if (!response.ok) {
+        return Promise.reject(new Error(`${response.status}`));
```
Volodymyr-Melnyk-Fron10der, line 21: I think it's better to throw the error here, not just reject

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,88 @@
...
+  const dispatch = useAppDispatch();
+
+  useEffect(() => {
+    try {
+      if (currentTodo?.userId) {
+        getUser(currentTodo.userId)
+          .then(setUser);
+      }
+    } catch (error) {
+      throw new Error(`User: ${currentTodo?.userId} not found. Error: ${error}`);
```
Volodymyr-Melnyk-Fron10der, line 20: do not throw it, just set in the state and show on the UI

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,88 @@
...
+    }
+  }, [currentTodo]);
+
+  if (!currentTodo) {
+    return (
+      <p>No todo found</p>
+    );
+  }
```
Volodymyr-Melnyk-Fron10der, line 28: move  on the 41 line

Volodymyr-Melnyk-Fron10der (changes requested):

src/components/TodoModal/TodoModal.tsx
Mykola-Hadupiak: i cant

Volodymyr-Melnyk-Fron10der: all returns should be at the end of the component. 
but ok, let it be

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,125 @@
...
+        return todos.filter(({ completed }) => !completed);
+      }
+
+      default:
+        return todosCopy;
+    }
+  }, [filterStatus]);
+
+  const todosForRender = useMemo(() => {
+    const todosCopy = [...filterTodosByStatus];
```
Volodymyr-Melnyk-Fron10der, line 32: you don't have to make a copy. filter is non-mutative method

Volodymyr-Melnyk-Fron10der (approved):

loralevitska (approved):
Well done)

PR #463, opened 2023-04-10
try to use redux with list of todos

OlhaYevstifieieva:


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,14 +1,25 @@
...
       className="field has-addons"
       onSubmit={event => event.preventDefault()}
     >
       <p className="control">
         <span className="select">
-          <select data-cy="statusSelect">
+          <select
+            data-cy="statusSelect"
+            onChange={(event) => {
+              dispatch(filterActions.setStatus(event.target.value as Status));
```
qoa11a, line 20: It's better to move these callbacks into separate functions and use them inside your attributes.

src/components/TodoList/TodoList.tsx
qoa11a, line 1: This component should be presentational. It means that you should pass todos as a prop here and render what's passed rather than getting it directly from the state.

qoa11a (changes requested):
Good job! 👍 
Just a few suggestions.

PR #466, opened 2023-04-13
solution

imariash86:
[DEMO LINK](https://imariash86.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,210 +1,109 @@
 /* eslint-disable max-len */
 import React from 'react';
+import cn from 'classnames';
+import { useDispatch } from 'react-redux';
+import { useAppSelector } from '../../app/hooks';
+import { actions } from '../../features/currentTodo';
 
 export const TodoList: React.FC = () => {
+  const todos = useAppSelector(state => state.todos);
+  const { query, status } = useAppSelector(state => state.filter);
+  const currentTodo = useAppSelector(state => state.currentTodo);
+  const dispatch = useDispatch();
+
+  let visibleTodos = [...todos];
+
+  if (query) {
+    visibleTodos = visibleTodos.filter(todo => todo.title.toLowerCase().includes(query.trim().toLowerCase()));
+  }
+
+  if (status !== 'all') {
+    visibleTodos = visibleTodos.filter(todo => {
+      switch (status) {
+        case 'active':
+          return !todo.completed;
+        case 'completed':
+          return todo.completed;
+        default:
+          throw new Error('Wrong filter');
+      }
+    });
+  }
+
   return (
     <>
-      <p className="notification is-warning">
-        There are no todos matching current filter criteria
-      </p>
+      {!visibleTodos.length
+        ? (
+          <p className="notification is-warning">
+            There are no todos matching current filter criteria
+          </p>
+        ) : (
+          <table className="table is-narrow is-fullwidth">
+            <thead>
+              <tr>
+                <th>#</th>
 
-      <table className="table is-narrow is-fullwidth">
-        <thead>
-          <tr>
-            <th>#</th>
+                <th>
+                  <span className="icon">
+                    <i className="fas fa-check" />
+                  </span>
+                </th>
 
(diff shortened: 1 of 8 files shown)


src/features/filter.ts
```diff
@@ -1,10 +1,55 @@
...
+  payload: status,
+});
+
+export const actions = { setQuery, clearQuery, setStatus };
+
+type Action = SetQueryAction | ClearQueryAction | SetStatusAction;
+
+type State = {
+  query: string,
+  status: Status,
```
Anna-Tkachenko, line 25: Consider using a more descriptive type name for 'Status' like 'TodoStatus' to make the code more readable 📖

src/features/filter.ts
```diff
@@ -1,10 +1,55 @@
-export const actions = { /* put action creators here */};
+import { Status } from '../types/Status';
```
Anna-Tkachenko, line 1: Consider using a more descriptive import name for 'Status' like 'TodoStatus' to make the code more readable 📖

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,70 @@
...
+      if (todo) {
+        const data = await getUser(todo.userId);
+
+        setUser(data);
+        setIsLoading(false);
+      }
+    };
+
+    currUser();
+  }, []);
```
Anna-Tkachenko, line 26: The useEffect dependency array is empty, which means it will only run once on the initial render. However, it should also run whenever the `todo` changes. Add `todo` to the dependency array.

src/features/todos.ts
```diff
@@ -1,9 +1,28 @@
 import { Todo } from '../types/Todo';
 
-export const actions = {};
+type SetAction = { type: 'todos/SET', payload: Todo[] };
+type AddAction = { type: 'todos/ADD', payload: Todo };
+type RemoveAction = { type: 'todos/REMOVE', payload: Todo };
```
Anna-Tkachenko, line 5: It's recommended to use the `typeof` keyword when defining action types to ensure consistency and avoid errors. For example, you can declare a constant like `const REMOVE_TODO = 'todos/REMOVE'` and then use `typeof REMOVE_TODO` as the type for the action.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,109 @@
...
+  const { query, status } = useAppSelector(state => state.filter);
+  const currentTodo = useAppSelector(state => state.currentTodo);
+  const dispatch = useDispatch();
+
+  let visibleTodos = [...todos];
+
+  if (query) {
+    visibleTodos = visibleTodos.filter(todo => todo.title.toLowerCase().includes(query.trim().toLowerCase()));
+  }
+
```
Anna-Tkachenko, line 19: will be better to create separate function for filtering and sorting todos

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,109 @@
...
+  if (status !== 'all') {
+    visibleTodos = visibleTodos.filter(todo => {
+      switch (status) {
+        case 'active':
```
Anna-Tkachenko, line 23: Let's create enum for `status`

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,70 @@
...
+    const currUser = async () => {
+      if (todo) {
+        const data = await getUser(todo.userId);
+
+        setUser(data);
+        setIsLoading(false);
+      }
+    };
+
+    currUser();
```
Anna-Tkachenko, line 25: Check function name, it's not clear what `currUser` function should do

src/features/currentTodo.ts
```diff
@@ -29,6 +29,11 @@ const currentTodoReducer = (
 ): State => {
   switch (action.type) {
     // Implement all actions here
+    case 'currentTodo/SET':
```
Anna-Tkachenko, line 32: will be better to use enum here

Anna-Tkachenko (changes requested):

PR #467, opened 2023-04-13
first commit

NVyrsta:


src/App.tsx
```diff
@@ -1,14 +1,43 @@
...
+    }
+  };
+
+  useEffect(() => {
+    loadTodosFromServer();
+  }, []);
+
+  const visibleTodos = filterTodos(todos, currentFilters.status, currentFilters.query);
+
+  const isLoadingFinished = (isLoadingError && !todos.length) || todos.length;
```
qoa11a, line 39: It's better to create a separate state `isLoading`.

qoa11a (changes requested):
See comments.

PR #468, opened 2023-04-13
Added React Redux List of Todos

ctk12:
[DEMO LINK](https://ctk12.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,34 +1,61 @@
...
+  const currentTodo = useAppSelector(state => state.currentTodo);
+
+  const loadTodos = async () => {
+    setIsLoading(true);
+    try {
+      const todosFromServer = await getTodos();
+
+      dispatch(actions.setTodos(todosFromServer));
+    } catch {
+      Promise.reject();
```
agniya-i, line 26: It's better to handle the error instead of just rejecting the promise. You can display an error message or handle it in any other appropriate way. 🚀

agniya-i (changes requested):
Please fix demo link, it's not working now

src/App.tsx
```diff
@@ -1,34 +1,59 @@
...
+import { actions } from './features/todos';
 
 export const App: React.FC = () => {
+  const [isLoading, setIsLoading] = React.useState(false);
+  const dispatch = useAppDispatch();
+  const currentTodo = useAppSelector(state => state.currentTodo);
+
+  const loadTodos = async () => {
+    setIsLoading(true);
+    try {
```
witflash, line 21: You have to use `catch` block if you use `try`. You can handle errors inside it (on this stage you can do `console.warn` for example).

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,102 @@
...
+              </div>
 
-      <Loader />
+              <button
+                type="button"
+                className="delete"
+                data-cy="modal-close"
+                onClick={handleRemoveTodo}
+              >
+                {null}
```
witflash, line 63: Why did you pass `null` inside button? Because of the linter?

witflash (approved):
Well done 👍🏻 
I left some comments to improve your future work.

PR #472, opened 2023-04-19
solved task

Serghii29:


src/App.tsx
```diff
@@ -21,14 +46,18 @@ export const App: React.FC = () => {
...
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isTodoLoading
+                ? (
+                  <Loader />
+                ) : (
+                  <TodoList />
+                )}
```
qoa11a, line 54: This way your `<TodoList />` will be shown even if there are errors.

src/components/TodoList/TodoList.tsx
qoa11a, line 1: This component should be presentational. It means it shouldn't get todos directly from the state. It should accept the todos it needs to render as a prop.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,90 @@
...
+    try {
+      if (currentTodo?.userId) {
+        setIscurrentTodoLoading(true);
+        const user = await getUser(currentTodo?.userId);
+
+        setUser(user);
+      }
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error('Error when receiving user:', error);
```
qoa11a, line 26: Handle errors.

src/features/filter.ts
```diff
@@ -1,10 +1,33 @@
-export const actions = { /* put action creators here */};
+type Action = {
+  type: 'filter/FILTER_QUERY' | 'filter/FILTER_STATUS';
+  payload: string;
+};
```
qoa11a, line 4: Create two separate types for filterQuery and filterStatus.

qoa11a (changes requested):
image
Some of the tests failed. Try to fix it.
Also, see other comments.

PR #476, opened 2023-04-20
result

svitjojo:


src/App.tsx
```diff
@@ -1,34 +1,61 @@
...
             <div className="block">
               <TodoFilter />
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {hasError
+                ? <span> No todos from server</span>
+                : <TodoList />}
```
qoa11a, line 52: This way your `<TodoList />` will be shown even if `loading` is true. I don't think it's a correct behaviour.

src/App.tsx
```diff
@@ -1,34 +1,61 @@
...
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { useAppDispatch, useAppSelector } from './app/hooks';
+import { todosActions } from './features/todos';
 
 export const App: React.FC = () => {
+  const [loading, setLoading] = useState(true);
```
qoa11a, line 15:
```suggestion
  const [isLoading, setIsLoading] = useState(true);
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,17 +1,35 @@
...
+    );
+  }, []);
+
+  const handleInput = useCallback((event: ChangeEvent<HTMLInputElement>) => {
+    dispatch(filterActions.setQuery(event.target.value));
+  }, []);
+
+  const handleClearQuery = useCallback(() => {
+    dispatch(filterActions.clearQuery());
+  }, []);
```
qoa11a, line 20: Don't use `useCallback` without a reason. In this case, it even worsens the performance a little bit.

You can read more about when and when not to use it:
https://react.dev/reference/react/useCallback

src/components/TodoList/TodoList.tsx
qoa11a, line 1: Don't access todos here. This component should be presentational.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,13 +1,30 @@
...
+import React, { useMemo } from 'react';
+import cn from 'classnames';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { currentTodosActions } from '../../features/currentTodo';
+import { getVisibleTodos } from '../../utils/getVisibleTodos';
 
 export const TodoList: React.FC = () => {
-  return (
-    <>
+  const { todos, currentTodo } = useAppSelector(state => state);
```
qoa11a, line 9: Don't access parts of the state like this. It will cause your component to re-render on every state change.

src/components/TodoModal/TodoModal.tsx
qoa11a, line 1: Pass currentTodo as a prop. This component should be presentational. I recommend you read more about presentational components.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,89 @@
...
+import { currentTodosActions } from '../../features/currentTodo';
+import { getUser } from '../../api';
+import { User } from '../../types/User';
 
 export const TodoModal: React.FC = () => {
+  const todo = useAppSelector(state => state.currentTodo);
+  const [user, setUser] = useState<User | null>(null);
+  const [isLoading, setIsLoading] = useState(true);
+  const [hasError, setHasError] = useState(false);
+  const [hasReceivedUser, setHasReceivedUser] = useState(false);
```
qoa11a, line 13: Redundant state. You can just check if there's a user by doing this `!!user`.

qoa11a (changes requested):
See comments.

PR #485, opened 2023-05-03
solution

Malva37:


src/App.tsx
```diff
@@ -1,14 +1,46 @@
...
+  const { query, status } = useAppSelector(state => state.filter);
+  // let todos: Todo[] = [];
+
+  const getTodosFromServer = async () => {
+    setLoadingError(true);
+    try {
+      const todosFromServer = await getTodos();
+
+      dispatch(actions.setTodos(todosFromServer));
+      setLoadingError(false);
```
qoa11a, line 28: You do it in `finally`. No need to do it here.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,17 +1,65 @@
...
+      case 'active':
+        return dispatch(filterActions
+          .filterActive(Status.ACTIVE, queryFromView));
+      case 'completed':
+        return dispatch(filterActions
+          .filterCompleted(Status.COMPLETED, queryFromView));
+      default:
+        return dispatch(filterActions.filterAll(Status.ALL, query));
+    }
+  };
```
qoa11a, line 43: It's overkill. A lot of redundant code.

I'd suggest creating just two action creators: `filterStatus` and `filterQuery`.
They'd look like this: `const filterStatus = (status: Status): FilterStatusAction => ({ type: 'todos/STATUS', payload: status });` and `const filterQuery = (query: string): FilterQueryAction => ({ type: 'todos/QUERY', payload: query });`

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,99 @@
...
-              </span>
-            </td>
+type Props = {
+  todos: Todo[] | null;
+};
 
-            <td className="is-vcentered is-expanded">
-              <p className="has-text-success">illo est ratione doloremque quia maiores aut</p>
-            </td>
+export const TodoList: React.FC<Props> = ({ todos }) => {
```
qoa11a, line 12: Wrap this component with `React.memo` to avoid extra re-renders.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,77 @@
...
+import { User } from '../../types/User';
 import { Loader } from '../Loader';
 
 export const TodoModal: React.FC = () => {
-  return (
-    <div className="modal is-active" data-cy="modal">
-      <div className="modal-background" />
+  const [hasLoadingError, setLoadingError] = useState(false);
+  const [currentUser, setCurrentUser] = useState<User | null>(null);
+  const currentTodo = useAppSelector(state => state.currentTodo);
```
qoa11a, line 11: It's better if you get this value as a prop of the component just like you did in `<TodoList />`.

src/features/filter.ts
qoa11a, line 1: I've commented about this in `<TodoFilter />` component. You need to rework the logic of this reducer.

qoa11a (changes requested):
Read my suggestions.
Also, there's a little problem when a user opens the todo. It blinks a little bit before showing the loader. You could fix it if you set your `isLoading` state to the `true` by default.

PR #486, opened 2023-05-03
Solution 1

IVxVI:
[DEMO LINK](https://ivxvi.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoModal/TodoModal.tsx
+++ b/src/components/TodoModal/TodoModal.tsx
@@ -1,40 +1,93 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
+import classNames from 'classnames';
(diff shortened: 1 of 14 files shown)


src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,93 @@
...
+  } = todo;
+
+  useEffect(() => {
+    const fetchUser = async () => {
+      try {
+        const selectedUser = await getUser(userId);
+
+        setUser(selectedUser);
+      } catch (error) {
+        throw new Error('Error while selecting user');
```
anstsot, line 28: it would be better to set error somewhere and show the message. now it will just break the whole app

anstsot (approved):
Great job, approved!

PR #490, opened 2023-05-09
finished task

OlyaZaiets:
[DEMO LINK](https://Olya Zaiets.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,210 +1,131 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import classNames from 'classnames';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { actions as todosActions } from '../../features/todos';
+import { getTodos } from '../../api';
+import { actions as currentTodoActions } from '../../features/currentTodo';
+import { filterTodos } from '../../utils';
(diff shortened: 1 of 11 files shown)


src/components/TodoList/TodoList.tsx
qoa11a, line 1: I'd recommend making this component presentational. It means that it should accept a list of todos as a prop and just render it.

src/components/TodoModal/TodoModal.tsx
qoa11a, line 1: I'd recommend making this component presentational as well.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,91 @@
...
+        setIsLoading(false);
+      }
+    };
+
+    fetchUser();
+  }, []);
+
+  if (!currentTodo) {
+    return null;
+  }
```
qoa11a, line 38: You shouldn't render this component at all if there's no `currentTodo`. In your case, you waste resources to run code in `useEffect` and to get state from Redux, to declare some local state with `useState`. Just get `currentTodo` in the `<App />` and pass it here as a prop. But also render this component like this: `{currentTodo && <TodoModal currentTodo={currentTodo} />}`.

qoa11a (changes requested):
image

Let's try to pass all the tests. :)

Also, see all other comments.

PR #492, opened 2023-05-13
 react_redux-list-of-todos

rafmamedov:


src/features/todos.ts
```diff
@@ -1,9 +1,28 @@
 import { Todo } from '../types/Todo';
 
-export const actions = {};
+type SetTodosAction = { type: 'todos/SET', payload: Todo[] };
 
-const todosReducer = (): Todo[] => {
-  return [];
+type Action = SetTodosAction;
+type State = Todo[];
```
SanyaBratashchuk, line 6:
```suggestion
type State = {
  isLoading; boolean;
  hasError: boolean;
  data: Todo[];
};
```
create async action for query todos. You will be able to remove local isLoading and error state

SanyaBratashchuk (changes requested):

src/features/todos.ts
rafmamedov: removed local isLoading and error state, added these into the async action in App

TarasHoliuk (approved):

PR #505, opened 2023-06-02
React_redux-list-of-todos

SerhiiVovchenko:
[DEMO LINK](https://SerhiiVovchenko.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,27 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
+import { useDispatch } from 'react-redux';
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { useAppSelector } from './app/hooks';
+import { getTodos } from './api';
+import { setTodos } from './features/todos';
 
 export const App: React.FC = () => {
+  const dispatch = useDispatch();
+  const selectedTodo = useAppSelector(state => state.currentTodo);
+  const todos = useAppSelector(state => state.todos);
+
+  useEffect(() => {
+    getTodos().then(apiTodos => {
+      dispatch(setTodos(apiTodos));
+    });
+  }, []);
+
   return (
     <>
       <div className="section">
@@ -21,14 +34,17 @@ export const App: React.FC = () => {
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {!todos.length
+                ? <Loader />
+                : <TodoList />}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {selectedTodo && (
+        <TodoModal selectedTodo={selectedTodo} />
+      )}
     </>
   );
 };
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,6 +1,26 @@
 import React from 'react';
+import { useDispatch } from 'react-redux';
+import { useAppSelector } from '../../app/hooks';
+import { actions as filterActions } from '../../features/filter';
+import { Status } from '../../types/Status';
 
 export const TodoFilter: React.FC = () => {
+  const { query, status } = useAppSelector(state => state.filter);
+  const dispatch = useDispatch();
+
+  const handleSelect: React.ChangeEventHandler<HTMLSelectElement>
+    = (event: React.ChangeEvent<HTMLSelectElement>) => {
+      dispatch(filterActions.setFilterType(event.target.value as Status));
+    };
+
+  const clearInput = () => {
+    dispatch(filterActions.setQuery(''));
+  };
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    dispatch(filterActions.setQuery(event.target.value));
+  };
+
   return (
     <form
       className="field has-addons"
@@ -8,7 +28,11 @@ export const TodoFilter: React.FC = () => {
     >
       <p className="control">
         <span className="select">
-          <select data-cy="statusSelect">
+          <select
+            data-cy="statusSelect"
+            value={status}
+            onChange={handleSelect}
+          >
             <option value="all">All</option>
             <option value="active">Active</option>
             <option value="completed">Completed</option>
@@ -22,19 +46,24 @@ export const TodoFilter: React.FC = () => {
           type="text"
           className="input"
           placeholder="Search..."
+          value={query}
+          onChange={handleChange}
         />
         <span className="icon is-left">
           <i className="fas fa-magnifying-glass" />
         </span>
 
-        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-          <button
-            data-cy="clearSearchButton"
-            type="button"
-            className="delete"
-          />
-        </span>
+        {!!query.length && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+            <button
+              data-cy="clearSearchButton"
+              type="button"
+              className="delete"
+              onClick={clearInput}
+            />
+          </span>
+        )}
       </p>
     </form>
   );
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,210 +1,102 @@
-/* eslint-disable max-len */
 import React from 'react';
+import classNames from 'classnames';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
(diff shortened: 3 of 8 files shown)


src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,76 @@
...
-      <div className="modal-card">
-        <header className="modal-card-head">
-          <div
-            className="modal-card-title has-text-weight-medium"
-            data-cy="modal-header"
-          >
-            Todo #3
-          </div>
+  useEffect(() => {
+    if (selectedTodo) {
```
Anastasiia-Svintsova, line 18: Consider handling possible errors when fetching the user data with a .catch() block.

src/features/todos.ts
```diff
@@ -1,9 +1,24 @@
...
+  { type: SET_TODOS, playload: todos }
+);
+
+const todosReducer = (
+  state: Todo[] = [],
+  action: SetTodos,
+): Todo[] => {
+  switch (action.type) {
+    case SET_TODOS:
+      return action.playload;
```
Anastasiia-Svintsova, line 17: Correct the typo 'playload' to 'payload'.

src/features/todos.ts
```diff
@@ -1,9 +1,24 @@
 import { Todo } from '../types/Todo';
 
-export const actions = {};
+const SET_TODOS = 'setTodos';
 
-const todosReducer = (): Todo[] => {
-  return [];
+type SetTodos = { type: 'setTodos', playload: Todo[] };
```
Anastasiia-Svintsova, line 5: Correct the typo 'playload' to 'payload'.

src/features/todos.ts
```diff
@@ -1,9 +1,24 @@
...
+
+export const setTodos = (todos: Todo[]) => (
+  { type: SET_TODOS, playload: todos }
```
Anastasiia-Svintsova, line 8: Correct the typo 'playload' to 'payload'.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,102 @@
...
-
-          <tr data-cy="todo">
-            <td className="is-vcentered">5</td>
-            <td className="is-vcentered"> </td>
-
-            <td className="is-vcentered is-expanded">
-              <p className="has-text-danger">laboriosam mollitia et enim quasi adipisci quia provident illum</p>
-            </td>
+  const dispatch = useAppDispatch();
+  const todos = useAppSelector(state => state.todos);
```
Anastasiia-Svintsova, line 9: It would be better to destructure the state object to make the code more readable. For example, 'const { todos, currentTodo, filter } = useAppSelector(state => state);'

src/App.tsx
```diff
@@ -21,14 +34,17 @@ export const App: React.FC = () => {
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {!todos.length
+                ? <Loader />
```
Anastasiia-Svintsova, line 38: Hm🤔 Let's think what will be if there is an empty array of todos from the server. You should add a state to check if it's loading or not

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,14 +1,38 @@
...
+import { useAppSelector } from '../../app/hooks';
+import { actions as filterActions } from '../../features/filter';
+import { Status } from '../../types/Status';
 
 export const TodoFilter: React.FC = () => {
+  const { query, status } = useAppSelector(state => state.filter);
+  const dispatch = useDispatch();
+
+  const handleSelect: React.ChangeEventHandler<HTMLSelectElement>
+    = (event: React.ChangeEvent<HTMLSelectElement>) => {
```
Anastasiia-Svintsova, line 12:
```suggestion
  const handleSelect: React.ChangeEventHandler<HTMLSelectElement> = (
    event: React.ChangeEvent<HTMLSelectElement>
  ) => {
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,14 +1,38 @@
...
+      dispatch(filterActions.setFilterType(event.target.value as Status));
+    };
+
+  const clearInput = () => {
+    dispatch(filterActions.setQuery(''));
```
Anastasiia-Svintsova, line 17: Use destructuring for the filterActions

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,14 +1,38 @@
...
         <span className="select">
-          <select data-cy="statusSelect">
+          <select
+            data-cy="statusSelect"
+            value={status}
+            onChange={handleSelect}
+          >
             <option value="all">All</option>
             <option value="active">Active</option>
             <option value="completed">Completed</option>
```
Anastasiia-Svintsova, line 38: Use an enum for the filter types

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -22,19 +46,24 @@ export const TodoFilter: React.FC = () => {
...
-          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-          <button
-            data-cy="clearSearchButton"
-            type="button"
-            className="delete"
-          />
-        </span>
+        {!!query.length && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
Anastasiia-Svintsova, line 58: Don't disable eslint, try to solve the error

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,102 @@
...
+      default:
+        return todos;
+    }
+  };
 
-          <tr data-cy="todo">
-            <td className="is-vcentered">8</td>
+  const list = selectList().filter(l => {
+    return l.title.toLowerCase().includes(query);
+  });
```
Anastasiia-Svintsova, line 32: Use memoization for expensive calculations

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,102 @@
...
 
             <td className="is-vcentered is-expanded">
-              <p className="has-text-success">illo est ratione doloremque quia maiores aut</p>
+              <p
+                className={classNames({
+                  'has-text-danger': !todo.completed,
+                },
+                {
+                  'has-text-success': todo.completed,
+                })}
```
Anastasiia-Svintsova, line 73: One of the possible options: 
```suggestion
                className={todo.completed ? 'has-text-success' : 'has-text-danger'}
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,76 @@
...
+          </header>
 
-            {/* For completed */}
-            <strong className="has-text-success">Done</strong>
-            {' by '}
-            <a href="mailto:Sincere@april.biz">Leanne Graham</a>
-          </p>
+          <div className="modal-card-body">
+            <p className="block" data-cy="modal-title">
+              {selectedTodo.title}
```
Anastasiia-Svintsova, line 55: Use destructuring for the selectedTodo

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,76 @@
...
+
+            <p className="block" data-cy="modal-user">
+              {selectedTodo.completed ? (
+                <strong className="has-text-success">Done</strong>
+              ) : (
+                <strong className="has-text-danger">Planned</strong>
+              )}
+              {' by '}
+              <a
+                href={`mailto:${currentUser.email}`}
```
Anastasiia-Svintsova, line 66: Same for the currentUser object

Anastasiia-Svintsova (changes requested):

OleziO (changes requested):
Fix filters.
image

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,102 @@
...
-            <td className="has-text-right is-vcentered">
-              <button data-cy="selectButton" className="button" type="button">
-                <span className="icon">
-                  <i className="far fa-eye" />
-                </span>
-              </button>
-            </td>
-          </tr>
+      default:
+        return todos;
```
TarasHoliuk, line 25:
```suggestion
        return true;
```
you don't actually return `todos` here - filter callback should return a boolean value, so returned by you `todos` is transformed to a boolean value under the hood - it is always true because it's an array. but better to return `true` explicitly (FYI)

src/App.tsx
```diff
@@ -1,14 +1,31 @@
...
 import { Loader } from './components/Loader';
+import { useAppSelector } from './app/hooks';
+import { getTodos } from './api';
+import { setTodos } from './features/todos';
 
 export const App: React.FC = () => {
+  const dispatch = useDispatch();
+  const selectedTodo = useAppSelector(state => state.currentTodo);
+  const todos = useAppSelector(state => state.todos);
+  const [isLoading, setIsLoading] = useState(false);
```
TarasHoliuk, line 18:
```suggestion
  const [isLoaded, setIsLoaded] = useState(false);
```
`isLoading` means that in the current moment, loading is in process, but according to cases you use this value, you store `true` if loading is done, so it should be named properly

TarasHoliuk (approved):
I've added a few comments just FYI (no need to re-request review)

PR #516, opened 2023-06-18
Redux todos list has been implemented

Oleksandr-Antonenko:
- [DEMO LINK](https://Oleksandr-Antonenko.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,14 +1,48 @@
...
+    try {
+      const todosFromServer = await getTodos();
+
+      dispatch(actions.setTodos(todosFromServer));
+    } catch {
+      setIsError(true);
+    } finally {
+      setIsLoading(false);
+    }
+  };
```
demetrxx, line 36: Makes sense to move the function inside useEffect to avoid re-renders, as it is used only there

PR #529, opened 2023-07-03
Ready

sergik94:
https://sergik94.github.io/react_redux-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,61 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
+import { useAppDispatch, useAppSelector } from './app/hooks';
+import { getTodos } from './api';
+import { actions as todosActions } from './features/todos';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 
 export const App: React.FC = () => {
+  const todos = useAppSelector(state => state.todos);
+  const selectedTodo = useAppSelector(state => state.currentTodo);
+  const { query, status } = useAppSelector(state => state.filter);
+  const dispatch = useAppDispatch();
+  const [isLoading, setLoading] = useState(false);
+
+  async function getTodosFromServer() {
+    setLoading(true);
+
+    try {
+      const data = await getTodos();
+
+      dispatch(todosActions.load(data));
+    } catch (error) {
+      window.console.log('Error: ', error);
+    } finally {
+      setLoading(false);
+    }
+  }
+
+  const visibleTodos = useMemo(() => {
+    if (status === 'all' && query !== '') {
+      return todos?.filter(todo => todo.title.includes(query));
+    }
+
+    if (status === 'completed') {
+      return todos?.filter(todo => {
+        return todo.completed === true && todo.title.includes(query);
+      });
+    }
+
+    if (status === 'active') {
+      return todos?.filter(todo => {
+        return todo.completed === false && todo.title.includes(query);
+      });
+    }
+
+    return todos;
+  }, [query, status, todos]) || null;
+
+  useEffect(() => {
+    getTodosFromServer();
+  }, []);
+
   return (
     <>
       <div className="section">
@@ -21,14 +68,13 @@ export const App: React.FC = () => {
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading ? <Loader /> : <TodoList todos={visibleTodos} />}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {selectedTodo !== null && <TodoModal />}
     </>
   );
 };
--- a/src/api.ts
+++ b/src/api.ts
@@ -21,4 +21,4 @@ function get<T>(url: string): Promise<T> {
 
 export const getTodos = () => get<Todo[]>('/todos');
 
-export const getUser = (userId: number) => get<User>(`/users/${userId}`);
+export const getUser = (userId: number | null) => get<User>(`/users/${userId}`);
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -42,13 +42,15 @@ export const TodoFilter: React.FC = () => {
         </span>
 
         <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-          <button
-            data-cy="clearSearchButton"
-            type="button"
-            className="delete"
-            onClick={() => dispatch(filterActions.setQuery(''))}
-          />
+          {query.trim() !== '' && (
+            <button
+              aria-label="Mute"
+              data-cy="clearSearchButton"
+              type="button"
+              className="delete"
+              onClick={() => dispatch(filterActions.setQuery(''))}
+            />
+          )}
         </span>
       </p>
     </form>
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
 import React from 'react';
 import { useAppDispatch, useAppSelector } from '../../app/hooks';
 import { actions as filterActions } from '../../features/filter';
@@ -44,7 +45,6 @@ export const TodoFilter: React.FC = () => {
         <span className="icon is-right" style={{ pointerEvents: 'all' }}>
           {query.trim() !== '' && (
             <button
-              aria-label="Mute"
               data-cy="clearSearchButton"
               type="button"
               className="delete"
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,6 +1,18 @@
 import React from 'react';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { actions as filterActions } from '../../features/filter';
 
 export const TodoFilter: React.FC = () => {
+  const { query } = useAppSelector(state => state.filter);
+  const dispatch = useAppDispatch();
+  const handleSelector = (event: React.FormEvent<HTMLSelectElement>) => {
+    dispatch(filterActions.setStatus(event.currentTarget.value));
+  };
+
+  const handleInput = (event: React.FormEvent<HTMLInputElement>) => {
+    dispatch(filterActions.setQuery(event.currentTarget.value));
+  };
+
   return (
     <form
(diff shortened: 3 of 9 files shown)


src/api.ts
```diff
@@ -21,4 +21,4 @@ function get<T>(url: string): Promise<T> {
 
 export const getTodos = () => get<Todo[]>('/todos');
 
-export const getUser = (userId: number) => get<User>(`/users/${userId}`);
+export const getUser = (userId: number | null) => get<User>(`/users/${userId}`);
```
mykhalenych, line 24: The 'getUser' function may not behave as expected when 'userId' is null. The resulting URL will become '/users/null' which likely does not match any route on the server. Consider adding a guard clause to handle the case when 'userId' is null.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,14 +1,27 @@
...
+
   return (
     <form
       className="field has-addons"
       onSubmit={event => event.preventDefault()}
     >
       <p className="control">
         <span className="select">
-          <select data-cy="statusSelect">
+          <select data-cy="statusSelect" onClick={handleSelector}>
```
mykhalenych, line 24: The select input should use the 'onChange' event instead of 'onClick'.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,85 @@
...
+    } finally {
+      setLoading(false);
+    }
+  }
+
+  const closeModal = () => {
+    dispatch(currentTodoActions.removeTodo());
+  };
+
+  useEffect(() => {
```
mykhalenych, line 33: The 'useEffect' hook is missing dependencies. It currently runs only once like 'componentDidMount', but if 'getUsersFromServer' is expected to run again when certain props or state change, you should include those variables in the dependency array.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,85 @@
...
+
+  async function getUsersFromServer() {
+    setLoading(true);
+
+    try {
+      const user = await getUser(selectedTodo?.userId || null);
+
+      setSelectedUser(user);
+    } catch (error) {
+      window.console.log('Error: ', error);
```
mykhalenych, line 23: Error handling is done just by logging it to the console. You might want to consider more user-friendly error handling. For instance, you could store the error in the state and display a message to the user.

src/App.tsx
```diff
@@ -1,14 +1,61 @@
...
+      dispatch(todosActions.load(data));
+    } catch (error) {
+      window.console.log('Error: ', error);
+    } finally {
+      setLoading(false);
+    }
+  }
+
+  const visibleTodos = useMemo(() => {
+    if (status === 'all' && query !== '') {
```
mykhalenych, line 36: The condition in the if statement might not work as expected. It appears that you want to filter todos when the status is 'all' and there is a query. If that's the case, consider changing '&&' to '||' because the current condition will only return true if both conditions are met.

src/App.tsx
```diff
@@ -1,14 +1,61 @@
...
+
+  async function getTodosFromServer() {
+    setLoading(true);
+
+    try {
+      const data = await getTodos();
+
+      dispatch(todosActions.load(data));
+    } catch (error) {
+      window.console.log('Error: ', error);
```
mykhalenych, line 29: Using console.log is not a good practice for error handling. Consider using an error handling library or a custom error handling function.

src/App.tsx
```diff
@@ -1,14 +1,61 @@
...
+      setLoading(false);
+    }
+  }
+
+  const visibleTodos = useMemo(() => {
+    if (status === 'all' && query !== '') {
+      return todos?.filter(todo => todo.title.includes(query));
+    }
+
+    if (status === 'completed') {
```
mykhalenych, line 40: create enum with completed, all etc

src/App.tsx
```diff
@@ -1,14 +1,61 @@
...
+    }
+
+    if (status === 'active') {
+      return todos?.filter(todo => {
+        return todo.completed === false && todo.title.includes(query);
+      });
+    }
+
+    return todos;
+  }, [query, status, todos]) || null;
```
mykhalenych, line 53:
```suggestion
  }, [query, status, todos]);
```

src/App.tsx
```diff
@@ -21,14 +68,13 @@ export const App: React.FC = () => {
...
-              <Loader />
-              <TodoList />
+              {isLoading ? <Loader /> : <TodoList todos={visibleTodos} />}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {selectedTodo !== null && <TodoModal />}
```
mykhalenych, line 77:
```suggestion
      {selectedTodo && <TodoModal />}
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,85 @@
-import React from 'react';
+/* eslint-disable react/jsx-one-expression-per-line */
```
mykhalenych, line 1: delete it and fix eslint error

mykhalenych (changes requested):

src/App.tsx
sergik94: @mykhalenych  this condition is corrected

src/App.tsx
```diff
@@ -1,14 +1,64 @@
...
 import { Loader } from './components/Loader';
+import { StatusEnum } from './types/Status';
 
 export const App: React.FC = () => {
+  const todos = useAppSelector(state => state.todos);
+  const selectedTodo = useAppSelector(state => state.currentTodo);
+  const { query, status } = useAppSelector(state => state.filter);
+  const dispatch = useAppDispatch();
+  const [isLoading, setLoading] = useState(true);
+  const [isError, setError] = useState(false);
```
etojeDenys, line 21:
```suggestion
  const [isError, setIsError] = useState(false);
```

src/App.tsx
```diff
@@ -1,14 +1,64 @@
...
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { StatusEnum } from './types/Status';
 
 export const App: React.FC = () => {
+  const todos = useAppSelector(state => state.todos);
+  const selectedTodo = useAppSelector(state => state.currentTodo);
+  const { query, status } = useAppSelector(state => state.filter);
+  const dispatch = useAppDispatch();
+  const [isLoading, setLoading] = useState(true);
```
etojeDenys, line 20:
```suggestion
  const [isLoading, setIsLoading] = useState(true);
```

src/App.tsx
```diff
@@ -1,14 +1,64 @@
...
+      return todos?.filter(todo => {
+        return todo.completed === true && todo.title.includes(query);
+      });
+    }
+
+    if (status === StatusEnum.Active) {
+      return todos?.filter(todo => {
+        return todo.completed === false && todo.title.includes(query);
+      });
+    }
```
etojeDenys, line 53: you can rewrite it using switch\case

src/App.tsx
```diff
@@ -21,14 +71,19 @@ export const App: React.FC = () => {
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading && <Loader />}
+              {!isError && !isLoading && <TodoList todos={visibleTodos} />}
+              {isError && !isLoading && (
```
etojeDenys, line 76: it looks so complex, try to split this logic using the ternary operator
```
{isLoading ? <Loader /> : ...}
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -22,18 +35,22 @@ export const TodoFilter: React.FC = () => {
...
         </span>
 
         <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-          <button
-            data-cy="clearSearchButton"
-            type="button"
-            className="delete"
-          />
+          {query.trim() !== '' && (
```
etojeDenys, line 46:
```suggestion
          {query.trim() && (
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,90 @@
...
-                  <i className="far fa-eye" />
-                </span>
-              </button>
-            </td>
-          </tr>
-        </tbody>
-      </table>
+  return (
+    <>
+      {todos?.length === 0 && todos !== null ? (
```
etojeDenys, line 20:
```suggestion
      {todos?.length ? (
```
we already checked if it's the array using `?`

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,90 @@
...
+                </th>
+
+                <th>Title</th>
+                <th> </th>
+              </tr>
+            </thead>
+
+            <tbody>
+              {
+                todos?.map(todo => (
```
etojeDenys, line 44: in overall why do we need to check if todos exist, if we already specified what we expect in props type

etojeDenys: so todos cannot be undefined or null

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,90 @@
...
+                  <tr
+                    key={todo.id}
+                    data-cy="todo"
+                    className={todo.id === selectedTodo?.id
+                      ? 'has-background-info-light'
```
etojeDenys, line 49: better to use classNames package for such complex class names

src/App.tsx
sergik94: @etojeDenys I could do it when I used Todo[] | null. But, after removing option 'null' I can't do it anymore

src/App.tsx
```diff
@@ -1,14 +1,64 @@
...
+import { getTodos } from './api';
+import { actions as todosActions } from './features/todos';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { StatusEnum } from './types/Status';
 
 export const App: React.FC = () => {
+  const todos = useAppSelector(state => state.todos);
+  const selectedTodo = useAppSelector(state => state.currentTodo);
+  const { query, status } = useAppSelector(state => state.filter);
```
IvanFesenko, line 18: Create separate selectors - so you can reuse them
`const todosSelector = (state: State) => state.todos;`

```suggestion
  const todos = useAppSelector(todosSelector);
  const selectedTodo = useAppSelector(state => state.currentTodo);
  const { query, status } = useAppSelector(state => state.filter);
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -22,18 +35,22 @@ export const TodoFilter: React.FC = () => {
...
-            data-cy="clearSearchButton"
-            type="button"
-            className="delete"
-          />
+          {query.trim() && (
+            <button
+              data-cy="clearSearchButton"
+              type="button"
+              className="delete"
+              onClick={() => dispatch(filterActions.setQuery(''))}
```
IvanFesenko, line 51: Try to avoid using anonymous function inside JSX - cretae separate handler and use here reference for it.

IvanFesenko (approved):

PR #536, opened 2023-07-10
react_redux-list-of-todos

Viktorianeimesh:
[DEMO LINK](https://Viktorianeimesh.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,14 +1,33 @@
-import React from 'react';
+import { ChangeEvent } from 'react';
+import { useDispatch } from 'react-redux';
+import { useAppSelector } from '../../app/hooks';
+import { actions } from '../../features/filter';
 
-export const TodoFilter: React.FC = () => {
+export const TodoFilter = () => {
+  const dispatch = useDispatch();
+  const { query, sort } = useAppSelector(state => state.filter);
+
+  const handleSortChange = ((event: ChangeEvent<HTMLSelectElement>) => {
+    dispatch(actions.sortChange(event.target.value));
+  });
+
+  const handleQueryChange = (event: ChangeEvent<HTMLInputElement>) => {
+    dispatch(actions.queryAdd(event.target.value));
+  };
+
+  const handleQueryClear = () => {
+    dispatch(actions.queryClear());
+  };
+
   return (
-    <form
-      className="field has-addons"
-      onSubmit={event => event.preventDefault()}
-    >
+    <form className="field has-addons">
       <p className="control">
         <span className="select">
-          <select data-cy="statusSelect">
+          <select
+            value={sort}
+            onChange={handleSortChange}
+            data-cy="statusSelect"
+          >
             <option value="all">All</option>
             <option value="active">Active</option>
             <option value="completed">Completed</option>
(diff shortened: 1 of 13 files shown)


src/features/todos.ts
```diff
@@ -1,9 +1,24 @@
...
+type Action = LoadTodos;
+
+const loadTodos = (todos: Todo[]): Action => ({
+  type: 'todos/LOAD',
+  payload: todos,
+});
+
+export const actions = { loadTodos };
+
+const todosReducer = (todos: Todo[] = [] || null, action: Action) => {
```
lerastarynets, line 14: The initial state of todos is set to [] || null, which is essentially []. Null will never be used because [] is a truthy value in JavaScript. If you want to have null as a possible initial state, consider revising this.

src/features/todos.ts
```diff
@@ -1,9 +1,24 @@
...
+  switch (action.type) {
+    case 'todos/LOAD':
+      return [...todos, ...action.payload];
```
lerastarynets, line 17: The 'todos/LOAD' case is currently returning a new array that is a combination of the old todos and the new ones. If you're intending to replace the old todos with the new ones, then the return statement should be 'return action.payload;' instead.

src/types/Sort.ts
```diff
@@ -0,0 +1,5 @@
+export enum Sort {
+  all = 'all',
```
lerastarynets, line 2: Pls see [enum naming convention](https://michaelcurrin.github.io/dev-cheatsheets/cheatsheets/typescript/data-types/enums.html#:~:text=Naming%20convention,enum%20tend%20to%20be%20titlecase)

src/features/filter.ts
```diff
@@ -1,10 +1,46 @@
...
+action: Action) => {
+  switch (action.type) {
+    case 'filter/QUERYADD':
+      return { ...filter, query: action.payload };
+
+    case 'filter/QUERYCLEAR':
+      return { ...filter, query: '' };
+
+    case 'filter/SORTCHANGE':
+      return { ...filter, sort: action.payload as Sort };
```
lerastarynets, line 39: Avoid using 'as' for type casting. It's better to handle the type checking in the action creators to ensure type safety. Consider changing the payload type of 'SortChangeAction' to 'Sort' to avoid this.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,14 +1,33 @@
...
+    dispatch(actions.queryClear());
+  };
 
-export const TodoFilter: React.FC = () => {
   return (
-    <form
-      className="field has-addons"
-      onSubmit={event => event.preventDefault()}
-    >
+    <form className="field has-addons">
```
lerastarynets, line 23: The form submission event is missing. It should prevent the default form submission behavior. Add onSubmit={(event) => event.preventDefault()} to the form element.

lerastarynets (changes requested):
Nice start! Left few comments on how to improve your code
Also I think that some king of message should be shown if no users where found

Screenshot at Jul 10 17-47-07

src/features/filter.ts
```diff
@@ -1,10 +1,46 @@
...
+
+export const actions = { queryAdd, queryClear, sortChange };
+
+const filterReducer = (filter: Filter = {
+  query: '',
+  sort: SORT.All,
+},
+action: Action) => {
+  switch (action.type) {
+    case 'filter/QUERYADD':
```
polosanya, line 32: It would be great to use Enum instead of hardcoded values

polosanya (approved):
Good job!

PR #543, opened 2023-07-12
solution

MSA1991:
[DEMO LINK](https://msa1991.github.io/react_redux-list-of-todos/)


src/App.tsx
```diff
@@ -1,14 +1,77 @@
...
+  };
+
+  const handleClearCurrentTodo = () => {
+    dispatch(currentTodoActions.removeTodo());
+  };
+
+  const getFilteredTodos = () => {
+    let newTodos = [...todos];
+
+    if (status !== 'all') {
```
etojeDenys, line 54: let's create some enum for status

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -1,14 +1,77 @@
...
+  const handleClearCurrentTodo = () => {
+    dispatch(currentTodoActions.removeTodo());
+  };
+
+  const getFilteredTodos = () => {
+    let newTodos = [...todos];
+
+    if (status !== Status.All) {
+      newTodos = newTodos.filter(todo => (
+        status === 'active' ? !todo.completed : todo.completed
```
ivbutsykin, line 56:
```suggestion
        status === Status.Active ? !todo.completed : todo.completed
```

src/App.tsx
```diff
@@ -1,14 +1,77 @@
...
+    dispatch(filterActions.changeStatus(value as Status));
+  };
+
+  const handleSetQuery = (value: string) => {
+    dispatch(filterActions.setQuery(value));
+  };
+
+  const handleClearQuery = () => {
+    dispatch(filterActions.clearQuery());
+  };
```
ivbutsykin, line 41: What do you think about to move it directly in `<TodoFilter />`?

Looks like it can works and can simplify your App component

src/App.tsx
```diff
@@ -17,18 +80,36 @@ export const App: React.FC = () => {
             <h1 className="title">Todos:</h1>
 
             <div className="block">
-              <TodoFilter />
+              <TodoFilter
+                status={status}
+                query={query}
```
ivbutsykin, line 85: Also you can take this values from store. It discussable that better pass it as prop or use from store, but I relates on redux recommendations:
 https://redux.js.org/style-guide/#connect-more-components-to-read-data-from-the-store

ivbutsykin (approved):
Great job!

PR #550, opened 2023-07-18
Commit for fe-chat

MykolaButylkov:
[DEMO LINK](https://mykolabutylkov.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/features/todos.ts
+++ b/src/features/todos.ts
@@ -1,9 +1,26 @@
 import { Todo } from '../types/Todo';
 
-export const actions = {};
+type SetTodosAction = {
+  type: 'todos/SET_TODOS',
+  payload: Todo[]
+};
 
-const todosReducer = (): Todo[] => {
-  return [];
+type Action = SetTodosAction;
+
+export const setTodos = (todos: Todo[]): SetTodosAction => ({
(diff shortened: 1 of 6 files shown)


mykhalenych (changes requested):
https://github.com/mate-academy/react_redux-list-of-todos/assets/53228259/dcb63a39-7a95-4a18-9452-1bf36c6e35b6

your app isn't working correctly

src/features/todos.ts
```diff
@@ -1,9 +1,26 @@
...
 
-const todosReducer = (): Todo[] => {
-  return [];
+const todosReducer = (todos = [], action: Action): Todo[] => {
+  switch (action.type) {
+    case 'todos/SET_TODOS':
+      return [...action.payload];
+    default:
+      return [...todos];
+  }
```
IvanFesenko, line 23: do not rewrite state in dafault case - it way cause troubles
```suggestion
const todosReducer = (state= [], action: Action): Todo[] => {
  switch (action.type) {
    case 'todos/SET_TODOS':
      return [...action.payload];
    default:
      return state;
  }
```

IvanFesenko (approved):
Pay attention on comment

PR #553, opened 2023-07-21
Reack redux todos have been implemented

sergey-vorobei:
[DEMO LINK](https://sergey-vorobei.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,17 +1,48 @@
-import React from 'react';
+import React, { ChangeEvent } from 'react';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { filterActions } from '../../features/filter';
+import { SelectValue } from '../../types/SelectValues';
 
 export const TodoFilter: React.FC = () => {
+  const dispatch = useAppDispatch();
+  const setQuery = (setToQuery: string) => dispatch(
+    filterActions.setQuery(setToQuery),
+  );
+  const setStatus = (setToStatus: string) => dispatch(
+    filterActions.setStatus(setToStatus),
+  );
+  const clearQuery = () => dispatch(filterActions.remove());
+  const { query, status } = useAppSelector(state => state.filter);
+
+  const handleChangeQuery = (e:ChangeEvent<HTMLInputElement>) => {
+    e.preventDefault();
+
+    setQuery(e.target.value);
+  };
+
+  const handleChangeSelect = (e:ChangeEvent<HTMLSelectElement>) => {
+    e.preventDefault();
+
+    setStatus(e.target.value as SelectValue);
+  };
+
   return (
-    <form
-      className="field has-addons"
-      onSubmit={event => event.preventDefault()}
-    >
+    <form className="field has-addons">
       <p className="control">
         <span className="select">
-          <select data-cy="statusSelect">
-            <option value="all">All</option>
-            <option value="active">Active</option>
-            <option value="completed">Completed</option>
+          <select
+            data-cy="statusSelect"
+            onChange={handleChangeSelect}
+            value={status}
+          >
+            {Object.entries(SelectValue).map(([key, value]) => (
+              <option
+                key={value}
+                value={value}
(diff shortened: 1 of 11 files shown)


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -20,20 +51,28 @@ export const TodoFilter: React.FC = () => {
...
+          {query && (
+            <button
+              data-cy="clearSearchButton"
+              name="clearButton"
+              type="button"
+              className="delete"
+              onClick={clearQuery}
+            >
+              <></>
+            </button>
```
Anastasiia-Svintsova, line 74:
```suggestion
            />
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,122 @@
...
+    }
+
+    return todosFromServer.filter((todo:Todo) => todo.title
+      .toLowerCase()
+      .includes(queryLower));
+  };
+
+  const visibleTodos = useMemo(() => {
+    return todosFilter(filterForStatusTodo(), query);
+  }, [todos, query, status]);
```
Anastasiia-Svintsova, line 40: Make sure that the dependencies array is correct. You should pass todosFilter, and filterForStatusTodo there as well. And remove todos and status. Wrap those functions in their own useMemo() and add dependencies there. Those functions are not computationally expensive, but it's still reasonable to use them as part of the memoization process, as it's part of the filtering process.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,122 @@
...
+              </th>
+              <th>Title</th>
+
+              <th> </th>
+            </tr>
+          </thead>
+
+          <tbody>
+            {
+              visibleTodos.map((todo) => {
```
Anastasiia-Svintsova, line 67:
```suggestion
            {visibleTodos.map((todo) => {
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,122 @@
...
+                return (
+                  <tr
+                    key={todo.id}
+                    data-cy="todo"
+                    className={classNames({
+                      'has-background-info-light': currentTodo?.id === todo.id,
+                    })}
+                  >
+                    <td className="is-vcentered">{todo.id}</td>
```
Anastasiia-Svintsova, line 76: Use destructuring for objects for better code readability. And BTW it would be great to create a separate component for Todo.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,91 @@
...
+import React, { useEffect, useState } from 'react';
+import classNames from 'classnames';
 import { Loader } from '../Loader';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { getUser } from '../../api';
+import { User } from '../../types/User';
+import { TotoActions } from '../../features/currentTodo';
 
 export const TodoModal: React.FC = () => {
+  const [isLoader, setIsLoader] = useState<boolean>(false);
```
Anastasiia-Svintsova, line 10: To avoid the jumpings that I mentioned before, you can use `true` for the initial value here and set it to false only if the loading ends
```suggestion
  const [isLoader, setIsLoader] = useState<boolean>(true);
```

Anastasiia-Svintsova (changes requested):
- Try to get rid of this jump when opening a modal window

https://github.com/mate-academy/react_redux-list-of-todos/assets/99403198/378526a6-aedc-4f14-a738-62a8b76baef5

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,122 @@
...
+
+              <th> </th>
+            </tr>
+          </thead>
+
+          <tbody>
+            {visibleTodos.map((todo) => {
+              return (
+                <TodoItem todoItem={todo} key={todo.id} />
+                // <tr
```
agniya-i, line 70: Don't forget to remove commented code)

agniya-i (approved):

PR #554, opened 2023-07-22
add task solution

NikaNika12:
[DEMO LINK](https://NikaNika12.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,14 +1,68 @@
...
+
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    const loadTodos = async () => {
+      try {
+        const loadedTodos = await getTodos();
+
+        dispatch(actionsTodos.setTodo(loadedTodos));
+        setIsLoading(false);
```
sTorba24, line 29: do it in `finally`

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,81 @@
...
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    const loadUsers = async () => {
+      try {
+        if (currentTodo) {
+          const loadedUsers = await getUser(currentTodo?.userId);
+
+          setUser(loadedUsers);
+          setIsLoading(false);
```
sTorba24, line 21: same

sTorba24 (changes requested):
Well done! Lets make it perfect

image
don't think it should be visible here :)

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,82 @@
...
+    const loadUsers = async () => {
+      try {
+        if (currentTodo) {
+          const loadedUsers = await getUser(currentTodo?.userId);
+
+          setUser(loadedUsers);
+        }
+      } catch (error) {
+        if (error) {
+          throw new Error('Users can not be loaded');
```
TarasHoliuk, line 24: Just FYI: now you throw the new Error on the error that happened during fetching a user. But you don't handle this new error at all, so it will lead to runtime error - the site will break. The correct way would be to show some error message here, in the same way as in the App.tsx

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -22,18 +34,26 @@ export const TodoFilter: React.FC = () => {
           type="text"
           className="input"
           placeholder="Search..."
+          value={query}
+          onChange={e => dispatch(
+            filterActions.setQuery(e.target.value as SortType),
```
TarasHoliuk, line 39:
```suggestion
            filterActions.setQuery(e.target.value),
```
type casting is redundant here. setQuery should accept any string

Also, it's better to create a separate function for such handlers

TarasHoliuk (approved):
Everything works fine, previous comments were fixed - approved :heavy_check_mark: 
But I've added 2 comments FYI - check them please

src/components/TodoFilter/TodoFilter.tsx
NikaNika12: Argument of type 'string' is not assignable to parameter of type 'SortType'.ts(2345) поэтому я добавила уточнее типа

TarasHoliuk (approved):

PR #564, opened 2023-07-31
add task solution

diamat1997:
[DEMO LINK](https://diamat1997.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,14 +1,70 @@
...
+        const loadedTodos = await getTodos();
+
+        dispatch(actionTodos.SetAdd(loadedTodos));
+        setIsLoading(false);
+      } catch (error) {
+        if (error instanceof Error) {
+          setErrorMessage(error.message);
+        }
+
+        setIsLoading(false);
```
loralevitska, line 35: `setIsLoading(false); ` repeatable code can be moved to the [`finally` block](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch)

src/App.tsx
```diff
@@ -21,14 +77,18 @@ export const App: React.FC = () => {
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading && <Loader />}
+              <TodoList
+                errorMessage={errorMessage}
+                todos={visibleTodos}
+                // selectTodo={selectTodo}
```
loralevitska, line 84: Remove comment. Try it write obvious code that explain all we need without comments)

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,87 @@
...
+            </span>
+          </th>
+          <th>Title</th>
+          <th> </th>
+        </tr>
+      </thead>
+
+      <tbody>
+        {!errorMessage
+          ? todos.map(todo => (
```
loralevitska, line 44: Destructure `todo`

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,94 @@
...
+          setIsLoading(false);
+        }
+      } catch (error) {
+        if (error instanceof Error) {
+          setErrorMessage(error.message);
+        }
+
+        setIsLoading(false);
+      }
+    };
```
loralevitska, line 33: Move function expression out from `useEffect`

loralevitska (changes requested):
Almost done) Let's improve

PR #569, opened 2023-08-02
react_redux-list-of-todos

partnersinbahamas:
[DEMO LINK](https://<partnersinbahamas>.github.io/react_redux-list-of-todos/)


src/App.tsx
```diff
@@ -1,14 +1,67 @@
...
+
+  const lowerQuery = query.toLowerCase();
+
+  const getData = async () => {
+    try {
+      setIsLoader(true);
+      let loadedData = await getTodos();
+
+      switch (status) {
+        case 'all':
```
anastasiia-tilikina, line 30: It is better to consider `all` case as default in `switch` instead of making it explicit.

src/App.tsx
```diff
@@ -1,14 +1,67 @@
...
+      dispatch(todosActions.setTodos(loadedData));
+    } catch (error) {
+      throw Error('Field to load todo data');
+    } finally {
+      setIsLoader(false);
+    }
+  };
+
+  useEffect(() => {
+    getData();
```
anastasiia-tilikina, line 62: In your app every re-render you make a new request to the server to get data. It is an expensive approach.
It is better to load data once the app is loaded for the first time, set them to store, and then just filter them in your app depending on filter options. And then just render those `visibleTodos`

Try to implement this

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,44 @@
...
-                <span className="icon">
-                  <i className="far fa-eye" />
-                </span>
-              </button>
-            </td>
-          </tr>
-        </tbody>
-      </table>
+      {!todos.length ? (
+        <p className="notification is-warning">
```
anastasiia-tilikina, line 14: This message is also shown during initial render along with the loader since in this case we also don't have todos yet. But it should be shown only when there's no todos based on our search query

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,91 @@
...
+  const currentUser: User = useAppSelector(state => state.currentUser);
+
+  const loadUserData = async () => {
+    try {
+      setIsLoader(true);
+      const loadedDataUser = await getUser(userId);
+
+      dispatch(currentUserActions.setUser(loadedDataUser));
+    } catch (error) {
+      throw Error('Failed to load user data');
```
anastasiia-tilikina, line 30: here is the same situation with error message

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,91 @@
...
+                <strong className="has-text-danger">Planned</strong>
+              )}
 
-            {/* For completed */}
-            <strong className="has-text-success">Done</strong>
-            {' by '}
-            <a href="mailto:Sincere@april.biz">Leanne Graham</a>
-          </p>
+              {' by '}
+              <a href="mailto:Sincere@april.biz">
```
anastasiia-tilikina, line 82: email is the same for all users now
fix it

anastasiia-tilikina (changes requested):
Well done)
you figured out how store works!

Let's just fix some issues I mentioned in comments!

VitaliyBondarenko1982 (approved):
Well done.
Looks like app work as expected.

PR #574, opened 2023-08-06
solution

Kinderwagen:
[DEMO LINK](https://Kinderwagen.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,208 +1,100 @@
 /* eslint-disable max-len */
 import React from 'react';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { actions as curentTodoFeat } from '../../features/currentTodo';
+import { Todo } from '../../types/Todo';
+import { Status } from '../../types/Status';
 
 export const TodoList: React.FC = () => {
+  const dispatch = useAppDispatch();
+
+  const currentTodo = useAppSelector(state => state.currentTodo);
+  const filter = useAppSelector(state => state.filter);
+  const todos = useAppSelector(state => state.todos.filter(
+    (todo: Todo) => {
+      return todo.title.toLocaleLowerCase().includes(filter.query.toLocaleLowerCase());
+    },
+  ));
+
+  const isSelected = (id: number) => currentTodo?.id === id;
+
+  const visibleTodos = () => {
+    switch (filter.status) {
+      case Status.All:
+        return todos.filter((todo: Todo) => todo);
+      case Status.Active:
+        return todos.filter((todo: Todo) => !todo.completed);
+      case Status.Completed:
+        return todos.filter((todo: Todo) => todo.completed);
+      default:
+        return todos;
+    }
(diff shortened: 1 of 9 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,208 +1,100 @@
...
+    switch (filter.status) {
+      case Status.All:
+        return todos.filter((todo: Todo) => todo);
+      case Status.Active:
+        return todos.filter((todo: Todo) => !todo.completed);
+      case Status.Completed:
+        return todos.filter((todo: Todo) => todo.completed);
+      default:
+        return todos;
+    }
```
etojeDenys, line 31: better to cover it in useMemo and use as var instead of function

src/components/TodoList/TodoList.tsx
```diff
@@ -1,208 +1,100 @@
...
+
+                <td className="has-text-right is-vcentered">
+                  <button
+                    data-cy="selectButton"
+                    className="button"
+                    type="button"
+                    onClick={() => dispatch(curentTodoFeat.setTodo(todo))}
+                  >
+                    <span className="icon">
+                      <i className={`far ${isSelected(id) ? 'fa-eye-slash' : 'fa-eye'}`} />
```
etojeDenys, line 91: let's use classNames package for such complex classes

src/features/filter.ts
```diff
@@ -1,10 +1,49 @@
...
+type AllActions = { type: 'ALL', payload: string };
+type ActiveActions = { type: 'ACTIVE', payload: string };
+type CompletedActions = { type: 'COMPLETED', payload: string };
+type InputActions = { type: 'INPUT', payload: string };
+
+type Action = AllActions | ActiveActions | CompletedActions | InputActions;
+
+const all = (query: string): Action => ({ type: 'ALL', payload: query });
+const active = (query: string): Action => ({ type: 'ACTIVE', payload: query });
+const completed = (query: string): Action => (
```
etojeDenys, line 12: better to create one action instead of these three and pass in this action in payload status value

etojeDenys (changes requested):

src/components/TodoList/TodoList.tsx
```diff
@@ -1,208 +1,100 @@
...
+  const visibleTodos = useMemo(() => {
+    switch (filter.status) {
+      case Status.All:
+        return todos.filter((todo: Todo) => todo);
+      case Status.Active:
+        return todos.filter((todo: Todo) => !todo.completed);
+      case Status.Completed:
+        return todos.filter((todo: Todo) => todo.completed);
+      default:
+        return todos;
```
etojeDenys, line 31:
```suggestion
    switch (filter.status) {
      case Status.Active:
        return todos.filter((todo: Todo) => !todo.completed);
      case Status.Completed:
        return todos.filter((todo: Todo) => todo.completed);
      case Status.All:
      default:
        return todos;
```

etojeDenys (approved):
lgtm

PR #579, opened 2023-08-09
react_redux-list-of-todos is done

Anastasiya145:
[DEMO LINK](https://Anastasiya145.github.io/react_redux-list-of-todos/)


src/features/filter.ts
```diff
@@ -1,10 +1,63 @@
...
+const setQuery = (query: string): SetQueryAction => ({
+  type: 'filter/QUERY_SET',
+  payload: query,
+});
+
+const clearQuery = (): ClearQueryAction => ({
+  type: 'filter/QUERY_CLEAR',
+});
+
+const setStatus = (status: string): SetStatusAction => ({
```
polosanya, line 26: The `status` parameter of the `setStatus` function should use `Status` instead of `string` type.
Fix all cases

src/App.tsx
```diff
@@ -21,14 +49,14 @@ export const App: React.FC = () => {
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading && <Loader />}
+              {isTodosFound && <TodoList />}
```
polosanya, line 53: The condition 'isTodosFound' will always evaluate to true because 'todos' is an array and in JavaScript, an empty array is a truthy value. Consider checking the length of 'todos' array instead.

src/App.tsx
```diff
@@ -1,14 +1,42 @@
...
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { useAppDispatch, useAppSelector } from './app/hooks';
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
+import { todosActions } from './features/todos';
 
 export const App: React.FC = () => {
+  const [isLoading, setIsLoading] = useState(false);
+  const dispach = useAppDispatch();
```
polosanya, line 17:
```suggestion
  const dispatch = useAppDispatch();
```

polosanya (changes requested):
Good job! Let's improve your code a bit)

lerastarynets (approved):
You're making excellent progress!

PR #582, opened 2023-08-10
solution

gabriella135:
[DEMO LINK](https://gabriella135.github.io/react_redux-list-of-todos/)


maxim2310 (changes requested):
Almost there) Just one more thing needs to be fixed
not need to make request on every query and status change just make filter of current state

PR #588, opened 2023-08-14
Task resolved

read-my-code:
[DEMO LINK](https://read-my-code.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,11 +1,10 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useMemo } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { useDispatch } from 'react-redux';
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
-import { Todo } from './types/Todo';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 import { getTodos } from './api';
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,14 +1,55 @@
...
+import { Todo } from './types/Todo';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { useAppSelector } from './app/hooks';
+import { actions as todosActions } from './features/todos';
+import { Status } from './types/Status';
 
 export const App: React.FC = () => {
+  const [isTodosAreLoaded, setIsTodosAreLoaded] = useState(false);
```
loralevitska, line 17: Redundant boolean. You can check todos.length instead

src/App.tsx
```diff
@@ -1,14 +1,55 @@
...
+      todosToFilter = todosToFilter.filter(
+        todo => todo.title.toLowerCase().includes(query.toLowerCase().trim()),
+      );
+    }
+
+    if (todosStatus === Status.Active) {
+      todosToFilter = todosToFilter.filter(todo => !todo.completed);
+    } else if (todosStatus === Status.Completed) {
+      todosToFilter = todosToFilter.filter(todo => todo.completed);
+    }
```
loralevitska, line 38: Combine repeatable code and do checking in filter method

loralevitska (changes requested):
Almost done) Let's improve

src/App.tsx
read-my-code: А якщо з сервера прийде помилка ? Тоді в мене буде вічне завантаження.

src/features/todos.ts
```diff
@@ -1,9 +1,22 @@
...
+});
+
+const todosReducer = (todos: Todo[] = [], action: Action): Todo[] => {
+  switch (action.type) {
+    case 'todos/ADD':
+      return [...todos, ...action.payload];
+
+    default:
+      return todos;
+  }
```
etojeDenys, line 18: do we have the ability to create the new todos?

etojeDenys (changes requested):

src/features/todos.ts
read-my-code: Це не передбачено самим завданням.

src/App.tsx
loralevitska: if an error comes from the server, you will catch it and the array will remain empty as it was by default before

loralevitska: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch

loralevitska (approved):
Good job!
P.S. read about promise.catch (I left link in comment)

PR #593, opened 2023-08-20
Solution

artyomwhite:
[DEMO LINK](https://artyomwhite.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,208 +1,100 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useMemo } from 'react';
+import classNames from 'classnames';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { actions as currentTodoFeat } from '../../features/currentTodo';
+import { Todo } from '../../types/Todo';
+import { Status } from '../../types/Status';
 
 export const TodoList: React.FC = () => {
+  const dispatch = useAppDispatch();
+
+  const currentTodo = useAppSelector(state => state.currentTodo);
+  const filter = useAppSelector(state => state.filter);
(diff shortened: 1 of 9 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,208 +1,100 @@
...
+      return todo.title.toLocaleLowerCase().includes(filter.query.toLocaleLowerCase());
+    },
+  ));
+
+  const isSelected = (id: number) => currentTodo?.id === id;
+
+  const visibleTodos = useMemo(() => {
+    switch (filter.status) {
+      case Status.All:
+        return todos.filter((todo: Todo) => todo);
```
VitaliyBodnar, line 25: todos list doesn't include null or undefined, so I don't any sense in filtering todos on Status.All
for this case, default case is enough

src/types/Status.ts
```diff
@@ -1 +1,9 @@
-export type Status = 'all' | 'active' | 'completed';
+export enum Status {
+
+  All = 'all',
+
+  Active = 'active',
+
+  Completed = 'completed',
+
+}
```
VitaliyBodnar, line 9:
```suggestion
export enum Status {
  All = 'all',
  Active = 'active',
  Completed = 'completed',
}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,208 +1,100 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useMemo } from 'react';
+import classNames from 'classnames';
```
VitaliyBodnar, line 3: `classNames` is too long name, it can be just `cn`

VitaliyBodnar (changes requested):
Almost done!

PR #596, opened 2023-08-21
Solution

ifuride:
[DEMO LINK](https://ifuride.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,40 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
-
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
+import { useAppDispatch, useAppSelector } from './app/hooks';
+import { getTodos } from './api';
+import { actions as todosActions } from './features/todos';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Error } from './components/Error';
 
 export const App: React.FC = () => {
+  const dispatch = useAppDispatch();
+  const todos = useAppSelector(state => state.todos);
+  const selectedTodo = useAppSelector(state => state.currentTodo);
+  const [isLoading, setIsLoading] = useState(false);
+  const [errorMessage, setErrorMessage] = useState('');
+
+  const loadTodos = async () => {
+    try {
+      setIsLoading(true);
+      const todosFromServer = await getTodos();
+
+      dispatch(todosActions.setTodos(todosFromServer));
+    } catch (error) {
+      setErrorMessage('Unable to load todos');
+    } finally {
+      setIsLoading(false);
+    }
+  };
+
+  useEffect(() => {
+    loadTodos();
+  }, []);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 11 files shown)


src/features/filter.ts
```diff
@@ -1,10 +1,49 @@
...
+);
+
+export const actions = { setQuery, removeQuery, setStatus };
+
+type Filter = {
+  query: string,
+  status: Status,
+};
+
+const initiialFilter: Filter = {
```
lerastarynets, line 25: There is a typo in the variable name 'initiialFilter'. It should be 'initialFilter'

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -22,19 +42,24 @@ export const TodoFilter: React.FC = () => {
           type="text"
           className="input"
           placeholder="Search..."
+          value={query}
+          onChange={(e) => setQuery(e.target.value)}
```
lerastarynets, line 46: The `setQuery` function is called on every key press. This might cause performance issues if the list of TODOs is large. Consider using a debounced version of the function to limit the number of dispatches.

src/App.tsx
```diff
@@ -21,14 +47,21 @@ export const App: React.FC = () => {
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading && (
+                <Loader />
+              )}
+              {todos.length > 0 && (
```
lerastarynets, line 53: pls see [falsy values](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)
```suggestion
              {!!todos.length && (
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,65 @@
...
 export const TodoModal: React.FC = () => {
+  const dispatch = useAppDispatch();
+  const selectedTodo = useAppSelector(state => state.currentTodo);
+
+  const [user, setUser] = useState<User | null>();
+
+  useEffect(() => {
+    if (selectedTodo) {
+      getUser(selectedTodo.userId)
+        .then(setUser);
```
lerastarynets, line 17: Error handling is missing for the promise returned by 'getUser'. You should add a catch block to handle any errors that may occur.

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,83 @@
...
+  const loadUser = async () => {
+    try {
+      setIsLoading(true);
+      if (selectedTodo) {
+        const userFromServer = await getUser(selectedTodo.userId);
+
+        setUser(userFromServer);
+      }
+    } catch {
+      throw new Error();
```
TarasHoliuk, line 26: throwing a new (+ empty) error is not correct error handling
since the error you throw here will not be handled anywhere - it will lead to an error and page crush. You should show the `Error` component, etc (in the same way you do in the App.tsx)

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,83 @@
...
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { User } from '../../types/User';
+import { getUser } from '../../api';
+import { actions as currentTodoActions } from '../../features/currentTodo';
 
 export const TodoModal: React.FC = () => {
+  const dispatch = useAppDispatch();
+  const selectedTodo = useAppSelector(state => state.currentTodo);
+
+  const [user, setUser] = useState<User | null>();
```
TarasHoliuk, line 12:
```suggestion
  const [user, setUser] = useState<User | null>(null);
```

TarasHoliuk (changes requested):
GJ :+1: 
Let's improve error handling in the TodoModal component

TarasHoliuk (approved):

PR #600, opened 2023-08-28
add solution

growingananas:
[DEMO LINK](https://growingananas.github.io/react_redux-list-of-todos/)


src/App.tsx
```diff
@@ -1,14 +1,47 @@
...
+
+      filteredTodos = todosFilterdByQuery(filteredTodos, query);
+
+      dispatch(TodosActions.setTodos(filteredTodos));
+    } catch (error) {
+      throw new Error('Failed to load todos');
+    }
+  };
+
+  useEffect(() => {
```
amashutikov, line 41: Looks like you are downloading all todos from server on every query or status change. Wich is not good. Maybe its better to create some variable visibleTodos and store in it filtered todos.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,14 +1,35 @@
...
       <p className="control">
         <span className="select">
-          <select data-cy="statusSelect">
+          <select
+            data-cy="statusSelect"
+            onChange={handleStatusChange}
+          >
             <option value="all">All</option>
             <option value="active">Active</option>
             <option value="completed">Completed</option>
```
amashutikov, line 35: You can render this using your Status enum `{
Object.entries(Status).map(([key, value]) => (
              <option
                value={value}
                key={key}
              >
                {key}
              </option>
            ))}
`

amashutikov (approved):
Well done! 
Only one possible mistake, on my opinion.

StasSokolov (approved):
Great job! I'm approving it)

PR #617, opened 2023-09-09
Develop

yuran007:
[DEMO LINK](https://yuran007.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,29 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
+import { useDispatch } from 'react-redux';
+import { useAppSelector } from './app/hooks';
+import { actions as todosActions } from './features/todos';
+
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
-import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { TodoModal } from './components/TodoModal';
+
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const dispatch = useDispatch();
+  const currentTodo = useAppSelector(state => state.currentTodo);
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    getTodos()
+      .then(fetchedTodos => dispatch(todosActions.getTodos(fetchedTodos)))
+      .finally(() => setIsLoading(false));
+  }, []);
+
   return (
     <>
       <div className="section">
@@ -21,14 +36,13 @@ export const App: React.FC = () => {
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading ? <Loader /> : <TodoList />}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {currentTodo && <TodoModal />}
     </>
   );
 };
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,14 +1,44 @@
-import React from 'react';
+import { useDispatch } from 'react-redux';
+import { useAppSelector } from '../../app/hooks';
+import { actions as filterActions } from '../../features/filter';
+
+import { StatusTodos } from '../../types/StatusTodos';
+import { ShowTodos } from '../../types/ShowTodos';
 
 export const TodoFilter: React.FC = () => {
+  const dispatch = useDispatch();
+  const query = useAppSelector(state => state.filter.query);
+
+  const handleFilterStatus = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    switch (event.target.value) {
+      case StatusTodos.Active:
+        dispatch(filterActions.setStatus(ShowTodos.Active));
+        break;
+
+      case StatusTodos.Completed:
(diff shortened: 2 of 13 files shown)


src/App.tsx
```diff
@@ -1,14 +1,29 @@
...
+
+import { getTodos } from './api';
 
 export const App: React.FC = () => {
+  const dispatch = useDispatch();
+  const currentTodo = useAppSelector(state => state.currentTodo);
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    getTodos()
```
Anastasiia-Svintsova, line 22: There is no error handling for the `getTodos` API call. If the API request fails, the user will not be informed and the loading spinner will continue indefinitely.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,14 +1,44 @@
...
+        break;
+    }
+  };
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    dispatch(filterActions.setQuery(event.target.value));
+  };
+
+  const clearQuery = () => dispatch(filterActions.setQuery(null));
+
```
Anastasiia-Svintsova, line 33: You are setting the query to null, which is not a good practice. Instead, you should set the initial value to an empty string.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,14 +1,44 @@
...
+import { ShowTodos } from '../../types/ShowTodos';
 
 export const TodoFilter: React.FC = () => {
+  const dispatch = useDispatch();
+  const query = useAppSelector(state => state.filter.query);
+
+  const handleFilterStatus = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    switch (event.target.value) {
+      case StatusTodos.Active:
+        dispatch(filterActions.setStatus(ShowTodos.Active));
```
Anastasiia-Svintsova, line 15: You can destructure the action from the filterActions for better readability

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,14 +1,44 @@
...
+  };
+
+  const clearQuery = () => dispatch(filterActions.setQuery(null));
+
   return (
-    <form
-      className="field has-addons"
-      onSubmit={event => event.preventDefault()}
-    >
+    <form className="field has-addons">
```
Anastasiia-Svintsova, line 35: You should prevent the default form submission event to avoid page reload.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,116 @@
...
-          </tr>
-
-          <tr data-cy="todo">
-            <td className="is-vcentered">7</td>
-            <td className="is-vcentered"> </td>
+export const TodoList: React.FC = () => {
+  const dispatch = useDispatch();
+  const todos = useAppSelector(state => state.todos);
+  const selectedStatus = useAppSelector(state => state.filter.status);
+  const query = useAppSelector(state => state.filter.query);
```
Anastasiia-Svintsova, line 14: Don't forget about the destructuring 
```suggestion
  const { status, query } = useAppSelector(state => state.filter);
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,91 @@
...
 export const TodoModal: React.FC = () => {
+  const dispatch = useDispatch();
+  const currentTodo = useAppSelector(state => state.currentTodo);
+  const [isLoading, setIsLoading] = useState(true);
+  const [user, setUser] = useState<User | null>(null);
+  const clearTodo = () => dispatch(currentTodoActions.removeTodo());
+
+  let todoID;
+  let title;
+  let completed;
```
Anastasiia-Svintsova, line 22: It's not a good practice to use `let` in the component. Why didn't you use state for these purposes?

src/features/filter.ts
```diff
@@ -1,10 +1,56 @@
-export const actions = { /* put action creators here */};
+import { ShowTodos } from '../types/ShowTodos';
 
-const filterReducer = () => {
-  return {
-    query: '',
-    status: 'all',
-  };
+type SetQueryAction = {
+  type: 'filterQuery/SET';
```
Anastasiia-Svintsova, line 4: Create an enum for types to avoid errors connected with typos

Anastasiia-Svintsova (changes requested):

src/App.tsx
yuran007: Fix

src/components/TodoFilter/TodoFilter.tsx
yuran007: Refactor

src/components/TodoFilter/TodoFilter.tsx
yuran007: Done

src/components/TodoFilter/TodoFilter.tsx
yuran007: Fix

src/components/TodoList/TodoList.tsx
yuran007: Refactor

src/components/TodoModal/TodoModal.tsx
yuran007: Fix

src/features/filter.ts
yuran007: Refactor

VitaliyBondarenko1982 (approved):
Well done

PR #619, opened 2023-09-11
Develop

cebanoleksandr:
- DEMO LINK - https://cebanoleksandr.github.io/react_redux-list-of-todos/


src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+          {todo.title}
+        </p>
+      </td>
+
+      <td className="has-text-right is-vcentered">
+        <button
+          data-cy="selectButton"
+          className="button"
+          type="button"
+          onClick={() => selectTodoHandler(todo)}
```
mbruhler, line 55: Avoid creating new functions within the render method. This could lead to unnecessary re-renders as a new function is created every time component re-renders.

src/features/filter.ts
```diff
@@ -1,10 +1,57 @@
...
+    type: 'filter/SET_QUERY',
+    payload: query,
   };
 };
 
+export const actions = { setStatus, setQuery };
+
+const initialState = {
+  query: '',
+  status: 'all' as Status,
```
mbruhler, line 31: Avoid using direct type casting (as keyword) if possible. Consider handling the initial state in a more appropriate way or adjust the Status type if necessary.

src/features/filter.ts
```diff
@@ -1,10 +1,57 @@
...
+};
+
+type InitialState = {
```
mbruhler, line 34: Type name 'InitialState' does not follow common naming conventions. Consider changing it to 'FilterState' or 'FilterInitialState' to make it more descriptive.

src/features/filter.ts
```diff
@@ -1,10 +1,57 @@
-export const actions = { /* put action creators here */};
+import { Status } from '../types/Status';
 
-const filterReducer = () => {
+type SetStatus = {
+  type: 'filter/SET_STATUS',
```
mbruhler, line 4: It would be better to extract 'filter/SET_STATUS' and 'filter/SET_QUERY' as constants and use them in your actions and reducer.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,75 @@
...
-      <div className="modal-card">
-        <header className="modal-card-head">
-          <div
-            className="modal-card-title has-text-weight-medium"
-            data-cy="modal-header"
-          >
-            Todo #3
-          </div>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+            <button
```
mbruhler, line 44: The button element has no content. If the button should be invisible to sighted users but accessible to screen readers, add an 'aria-label' attribute to provide a label for screen readers.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,75 @@
...
+type Props = {
+  todo: Todo;
+};
+
+export const TodoModal: React.FC<Props> = ({ todo }) => {
+  const [user, setUser] = useState<User | null>(null);
+  const dispatch = useAppDispatch();
+
+  useEffect(() => {
+    getUser(todo?.userId).then(data => {
```
mbruhler, line 18: You are not handling the potential error from the 'getUser' promise. This can lead to unhandled promise rejections. You should add a '.catch()' block to handle potential errors.

src/App.tsx
```diff
@@ -1,14 +1,30 @@
...
+  const dispatch = useAppDispatch();
+
+  useEffect(() => {
+    setIsLoading(true);
+
+    getTodos().then(data => {
+      dispatch(todosActions.setTodos(data));
+      setIsLoading(false);
+    });
+  }, []);
```
mbruhler, line 26: You are missing the catch block for error handling in case the promise from 'getTodos' function rejects.

src/App.tsx
```diff
@@ -1,14 +1,30 @@
...
 
 export const App: React.FC = () => {
+  const [isLoading, setIsLoading] = useState(false);
+  const todo = useAppSelector(state => state.currentTodo);
+  const dispatch = useAppDispatch();
+
+  useEffect(() => {
+    setIsLoading(true);
+
+    getTodos().then(data => {
```
mbruhler, line 22: The response from the 'getTodos' function is not being checked before dispatching the action to set the todos. You should check if the data exists and handle any potential errors.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,17 +1,25 @@
...
+  const dispatch = useAppDispatch();
+  const { status, query } = useAppSelector(state => state.filter);
 
-export const TodoFilter: React.FC = () => {
   return (
-    <form
-      className="field has-addons"
-      onSubmit={event => event.preventDefault()}
-    >
+    <form className="field has-addons">
```
mbruhler, line 10: You should prevent the default form submission event to avoid page reload. Add 'onSubmit' event to the form and call 'event.preventDefault()' inside.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -22,18 +30,23 @@ export const TodoFilter: React.FC = () => {
...
 
         <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-          <button
-            data-cy="clearSearchButton"
-            type="button"
-            className="delete"
-          />
+          {!!query && (
+            /* eslint-disable-next-line jsx-a11y/control-has-associated-label */
```
mbruhler, line 42: You have a linting rule to always have a label associated with form control. Though you've disabled it for the button, it's better to have a visually hidden label for accessibility.

mbruhler (changes requested):
Good work on your side!
Remember to try-catch in `useEffect` hook, especially when fetching something from API.
Keep in mind also naming conventions.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,80 @@
...
-          >
-            Todo #3
-          </div>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+            <button
+              type="button"
+              className="delete"
+              data-cy="modal-close"
+              onClick={closeModal}
+              area-label
```
And678, line 53: This is a label for screen readers, you need to provide text there
```suggestion
              aria-label="Close"
```

src/features/filter.ts
```diff
@@ -1,10 +1,60 @@
...
+const SET_QUERY = 'filter/SET_QUERY';
+
+type SetStatus = {
+  type: 'filter/SET_STATUS',
+  payload: Status,
+};
+
+const setStatus = (status: Status): SetStatus => {
+  return {
+    type: 'filter/SET_STATUS',
```
And678, line 13:
```suggestion
    type: SET_STATUS,
```

src/features/filter.ts
```diff
@@ -1,10 +1,60 @@
-export const actions = { /* put action creators here */};
+import { Status } from '../types/Status';
 
-const filterReducer = () => {
+const SET_STATUS = 'filter/SET_STATUS';
+const SET_QUERY = 'filter/SET_QUERY';
+
+type SetStatus = {
+  type: 'filter/SET_STATUS',
```
And678, line 7:
```suggestion
  type: SET_STATUS,
```

Same for the rest of actions/reducers

And678 (changes requested):
Good work, only a couple of things to fix.

Firstly, check out the tests and lint, they are failing.

And678 (approved):
Good job, approved :white_check_mark:

PR #626, opened 2023-09-13
solution

MariiaMonakhova:
[DEMO LINK](https://MariiaMonakhova.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,14 +1,39 @@
 import React from 'react';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { actions as filterActions } from '../../features/filter';
+import { Status } from '../../types/Status';
 
 export const TodoFilter: React.FC = () => {
+  const { query, status } = useAppSelector(state => state.filter);
+  const dispatch = useAppDispatch();
+
+  const handleStatusChange = (todoStatus: Status) => {
+    return dispatch(filterActions.filterByStatus(todoStatus));
+  };
+
+  const handleQueryChange = (inputQuery: string) => {
+    return dispatch(filterActions.filterByQuery(inputQuery));
+  };
+
+  const clearQuery = () => {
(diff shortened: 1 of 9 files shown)


cebanoleksandr (approved):
GJ)

src/utils/filterTodos.ts
```diff
@@ -0,0 +1,31 @@
+import { Status } from '../types/Status';
+import { Todo } from '../types/Todo';
+
+export function filterTodos(todos: Todo[], query: string, status: Status) {
+  let preparedTodos = [...todos];
```
ArtemTeslenko, line 5: array is not a primitive data type, you can change it even if it will be a constant, no need to use let

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,14 +1,39 @@
...
+
+  const clearQuery = () => {
+    return dispatch(filterActions.clearQuery());
+  };
+
   return (
     <form
       className="field has-addons"
-      onSubmit={event => event.preventDefault()}
+      onSubmit={(event) => event.preventDefault()}
```
ArtemTeslenko, line 25: can leave without brackets

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,96 @@
...
-
-            <td className="has-text-right is-vcentered">
-              <button data-cy="selectButton" className="button" type="button">
-                <span className="icon">
-                  <i className="far fa-eye" />
-                </span>
-              </button>
-            </td>
-          </tr>
+  const filteredTodos = filterTodos(allTodos, query, status);
```
And678, line 16: Use `useMemo` here.

And678 (approved):
Approve :white_check_mark: , but read my comment below.

ArtemTeslenko (approved):
Good job. There are some comments, read and consider next time. Also, two hundred items is not a huge array, but you could use useMemo for practice)

PR #634, opened 2023-09-15
Add task solution

vladmakohon:
[DEMO LINK](https://vladmakohon.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,210 +1,94 @@
-/* eslint-disable max-len */
 import React from 'react';
+import classNames from 'classnames';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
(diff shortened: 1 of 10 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,94 @@
...
-
-            <td className="has-text-right is-vcentered">
-              <button data-cy="selectButton" className="button" type="button">
-                <span className="icon">
-                  <i className="far fa-eye" />
-                </span>
-              </button>
-            </td>
-          </tr>
+  const visibleTodos = getFilteredTodos(todos, filter.query, filter.status);
```
And678, line 13: Use `useMemo` here to avoid running this on rerenders

And678 (approved):
Approved, but check my comment

PR #647, opened 2023-09-25
added task solution

AnastasiiaKhemych:
[DEMO LINK](https://AnastasiiaKhemych.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,210 +1,99 @@
 /* eslint-disable max-len */
 import React from 'react';
+import cn from 'classnames';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { Todo } from '../../types/Todo';
+import { Status } from '../../types/Status';
+import { actions } from '../../features/currentTodo';
+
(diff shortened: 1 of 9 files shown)


vlopatko (approved):

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,99 @@
...
-                <span className="icon">
-                  <i className="far fa-eye" />
-                </span>
-              </button>
-            </td>
-          </tr>
+  const todos = useAppSelector(state => state.todos);
+  const selectedTodo = useAppSelector(state => state.currentTodo);
+  const { query, status } = useAppSelector(state => state.filter);
+  const preparedTodos = getNormalisedTodos(todos, query, status);
```
And678, line 34: Use `useMemo` here to avoid running this on every render (e.g when selected todo changes)

src/types/Status.ts
And678, line 1: Better call this file types.ts, because it has other types, not only Status.

And678 (approved):
Approved, but check my comment below

PR #682, opened 2023-11-01
task solution

UlianaChorna:
[DEMO LINK](https://UlianaChorna.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,6 +1,20 @@
 import React from 'react';
+import { useDispatch } from 'react-redux';
+import { useAppSelector } from '../../app/hooks';
+import { actions } from '../../features/filter';
 
 export const TodoFilter: React.FC = () => {
+  const dispatch = useDispatch();
+  const filter = useAppSelector(state => state.filter);
+
+  const handleStatusChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    dispatch(actions.addStatus(event.target.value));
+  };
+
+  const handleQuery = (query = '') => {
+    dispatch(actions.addQuery(query));
+  };
+
   return (
     <form
       className="field has-addons"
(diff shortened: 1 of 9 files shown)


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -22,19 +40,25 @@ export const TodoFilter: React.FC = () => {
...
-          />
-        </span>
+        {filter.query && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+            <button
+              data-cy="clearSearchButton"
+              type="button"
+              className="delete"
+              onClick={() => handleQuery()}
```
IvanFesenko, line 57: Woks same, but less callbacks
```suggestion
              onClick={handleQuery}
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,118 @@
...
+          return filterQuery;
+        case Status.ACTIVE:
+          return !todo.completed && filterQuery;
+        case Status.COMPLETED:
+          return todo.completed && filterQuery;
+        default:
+          return filterQuery;
+      }
+    });
+  };
```
IvanFesenko, line 34: No need to filter incase Status.ALL or query is empty

IvanFesenko (changes requested):
Looks good, fix minor comments

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,116 @@
...
+      switch (filter.status) {
+        case Status.ACTIVE:
+          return !todo.completed && filterQuery;
+        case Status.COMPLETED:
+          return todo.completed && filterQuery;
+        default:
+          return filterQuery;
+      }
+    });
+  };
```
IvanFesenko, line 32: Not fully understand comment. No need to call Array.filter() at all in case status is all

IvanFesenko (approved):

PR #690, opened 2023-11-12
solution

Gosyav:
https://vo7kov.github.io/react_redux-list-of-todos/

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,6 +1,21 @@
 import React from 'react';
+import { SortType } from '../../types/SortType';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { actions as filterActions } from '../../features/filter';
 
 export const TodoFilter: React.FC = () => {
+  const dispatch = useAppDispatch();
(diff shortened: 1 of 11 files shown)


DarkMistyRoom (changes requested):
Please, use lowercase for the filter's status values. Everything else is pretty good!
image

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,17 +1,36 @@
...
-          <select data-cy="statusSelect">
-            <option value="all">All</option>
-            <option value="active">Active</option>
-            <option value="completed">Completed</option>
+          <select
+            data-cy="statusSelect"
+            onChange={hadnleSelectChange}
+            value={sortType}
+          >
+            <option value={SortType.all}>All</option>
```
TarasHoliuk, line 31:
```suggestion
            <option value={SortType.All}>all</option>
```
PascalCase is ok for enum keys

The requested change was about text you are displaying :)

TarasHoliuk (approved):
Added 1 comment FYI - the previous comment wasn't fixed properly, but it wasn't critical to reject 2 times because of it, so approved (agree - code LGTM :+1: )

PR #692, opened 2023-11-13
develop

yuliabakun:
[demo](https://yulabakun.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,210 +1,88 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { TodoItem } from '../TodoItem/TodoItem';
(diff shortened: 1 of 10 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,88 @@
...
+        break;
+    }
+  }
+
+  return filteredTodos;
+}
+
+export const TodoList: React.FC<Props> = ({ todos }) => {
+  const query = useAppSelector(state => state.filter.query);
+  const filterByStatus = useAppSelector(state => state.filter.status);
```
vasylmiskiv, line 45: Also you could destructure the properties 
```suggestion
  const { query, status: filterByStatus } = useAppSelector(state => state.filter);
```

vasylmiskiv (approved):
Looks fine but pay attention to the comments please

PR #702, opened 2023-11-15
create redux logic

viktoria-roik:
[DEMO LINK](https://viktoria-roik.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,24 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { useAppDispatch, useAppSelector } from './app/hooks';
+import { getTodos } from './api';
+import { actions as setTodosActions } from './features/todos';
 
 export const App: React.FC = () => {
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+  const todos = useAppSelector(state => state.todos);
+  const currentTodo = useAppSelector(state => state.currentTodo);
+  const dispatch = useAppDispatch();
+
+  useEffect(() => {
+    setIsLoading(true);
+
+    getTodos()
+      .then((todoFromServer) => dispatch(setTodosActions.setTodos(todoFromServer)))
+      .finally(() => setIsLoading(false));
+  }, []);
+
   return (
     <>
       <div className="section">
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,14 +1,30 @@
...
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { useAppDispatch, useAppSelector } from './app/hooks';
+import { getTodos } from './api';
+import { actions as setTodosActions } from './features/todos';
 
 export const App: React.FC = () => {
+  const [isLoading, setIsLoading] = useState<boolean>(false);
```
rostyslav48, line 15: There is no need to pass generic here, because boolean is a simple value

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,14 +1,37 @@
...
       <p className="control">
         <span className="select">
-          <select data-cy="statusSelect">
+          <select
+            data-cy="statusSelect"
+            onChange={(event) => {
+              handleStatusChange(event.target.value as Status);
+            }}
+          >
             <option value="all">All</option>
```
rostyslav48, line 35: You can create enum for these values and then map them

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -22,18 +45,24 @@ export const TodoFilter: React.FC = () => {
...
         </span>
 
         <span className="icon is-right" style={{ pointerEvents: 'all' }}>
           {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-          <button
-            data-cy="clearSearchButton"
-            type="button"
-            className="delete"
-          />
+          {!!query.length && (
```
rostyslav48, line 57: You can simply use `query`, the result would be the same)

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,115 @@
...
-            <td className="has-text-right is-vcentered">
-              <button data-cy="selectButton" className="button" type="button">
-                <span className="icon">
-                  <i className="far fa-eye" />
-                </span>
-              </button>
-            </td>
-          </tr>
+    if (filter.query) {
+      newTodos = newTodos.filter(
```
rostyslav48, line 18: It is a good practice to move such function to a separate file

src/features/filter.ts
```diff
@@ -1,10 +1,50 @@
...
+  status: Status,
+};
+
+type Action = SetQueryAction | SetStatusAction;
+
+export const actions = { setQuery, setStatus };
+
+const InitialState = {
+  query: '',
+  status: 'all' as Status,
```
rostyslav48, line 25: You can use enum for completionStatus

rostyslav48 (approved):
Good Job😎.
There are some comments below to take into consideration)

PR #722, opened 2023-11-27
initial commit

cardinal1991:
[DEMO LINK](https://cardinal1991.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,26 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
-import { Loader } from './components/Loader';
+import { Loader } from './components/Loader/Loader';
+import { useAppDispatch, useAppSelector } from './app/hooks';
+import { fetchTodos } from './features/todos';
 
 export const App: React.FC = () => {
+  const selectedTodos = useAppSelector((state) => state.currentTodo);
+
+  const dispatch = useAppDispatch();
+
+  useEffect(() => {
+    dispatch(fetchTodos());
+  }, [dispatch]);
+  // any ze wzgl na konflikty w nazwach
+  const todos = useAppSelector((state: any) => state.todos.todos);
+
   return (
     <>
       <div className="section">
@@ -21,14 +33,15 @@ export const App: React.FC = () => {
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {(todos?.length > 0)
+                ? <TodoList todos={todos} />
+                : <Loader />}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {selectedTodos && <TodoModal />}
     </>
   );
 };
--- a/src/app/store.ts
+++ b/src/app/store.ts
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
 import { createStore, applyMiddleware, combineReducers } from 'redux';
 import { composeWithDevTools } from 'redux-devtools-extension';
 import thunk from 'redux-thunk';
@@ -19,6 +20,16 @@ export const store = createStore(
     applyMiddleware(thunk),
   ),
 );
+// to nam pozwala na sledzenie zmian w storze
+const history: any = [];
+
+store.subscribe(() => {
+  const state = store.getState();
+
+  // console.log('state', state);
+  history.push(state);
+  console.log('history', history);
+});
 
 export type RootState = ReturnType<typeof store.getState>;
 export type AppDispatch = typeof store.dispatch;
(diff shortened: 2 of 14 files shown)


src/App.tsx
```diff
@@ -1,14 +1,26 @@
...
 export const App: React.FC = () => {
+  const selectedTodos = useAppSelector((state) => state.currentTodo);
+
+  const dispatch = useAppDispatch();
+
+  useEffect(() => {
+    dispatch(fetchTodos());
+  }, [dispatch]);
+  // any ze wzgl na konflikty w nazwach
+  const todos = useAppSelector((state: any) => state.todos.todos);
```
choeqq, line 22: instead of using it as any just change the names 😄

src/App.tsx
```diff
@@ -21,14 +33,15 @@ export const App: React.FC = () => {
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {(todos?.length > 0)
+                ? <TodoList todos={todos} />
+                : <Loader />}
```
choeqq, line 38: if there will be no data due to an error, the loading state will persist till the error is solved, please provide some kind of loading state and base your ternary upon that

src/app/store.ts
```diff
@@ -19,6 +20,16 @@ export const store = createStore(
     applyMiddleware(thunk),
   ),
 );
+// to nam pozwala na sledzenie zmian w storze
+const history: any = [];
```
choeqq, line 24: please try to find a better way to go about it, consider using generics 😉

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -22,18 +32,33 @@ export const TodoFilter: React.FC = () => {
...
           placeholder="Search..."
+          value={searchText}
+          onChange={(e) => {
+            dispatch(actions.setSearchText(e.target.value));
+            if (e.target.value !== '') {
+              setIsClear(true);
+            } else {
+              setIsClear(false);
+            }
+          }}
```
choeqq, line 43: I think this function should be extracted from JSX

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -22,18 +32,33 @@ export const TodoFilter: React.FC = () => {
...
+              type="button"
+              className="delete"
+              onClick={() => {
+                dispatch(actions.setSearchText(''));
+                setIsClear(false);
+              }}
+              aria-label="Clear Search"
+            />
+          )}
         </span>
```
choeqq, line 62: I am not a fan of this setup, why is span wrapping a button? and why button does not have content?

src/components/TodoList/TodoList.tsx
```diff
@@ -1,12 +1,27 @@
 /* eslint-disable max-len */
 import React from 'react';
+import { useAppSelector } from '../../app/hooks';
+import { Todo } from '../Todo/todo';
+import { filterTodos } from '../filteredTodos/filteredTodos';
+
+type Props = {
+  // wygodniej dac any bo nazwa typu Todo pokrywa sie z nazwa komponentu Todo
+  todos: any[];
```
choeqq, line 9: again, if you're typing everything as any, using TS has basically no point, I know it's convenient in some particular cases, but most of the time there are many workarounds

src/components/filteredTodos/filteredTodos.tsx
```diff
@@ -0,0 +1,28 @@
...
+      case 'completed':
+        return todo.completed
+        && todo.title.toLowerCase().includes(searchText.toLowerCase().trim());
+      case 'active':
+        return !todo.completed
+        && todo.title.toLowerCase().includes(searchText.toLowerCase().trim());
+      default:
+        return true;
+    }
+  });
```
choeqq, line 25: imo this function belongs to the redux store

src/context/Context.tsx
choeqq, line 1: why are you using context when the whole point of Redux is to provide you with the global store?
you should use either Redux OR Context, they can work together, but since the whole point of this project is to learn Redux I think you should stay with Redux store

src/features/filter.ts
```diff
@@ -1,10 +1,49 @@
...
+//   query: string;
+//   status: Filter;
+// }
+
+const initialState = {
+  query: '',
+  status: Filter.All,
+};
+
+const filterReducer = (state = initialState, action: any) => {
```
choeqq, line 32: action will probably `keyof typeof actions`

src/features/filter.ts
```diff
@@ -1,10 +1,49 @@
...
+
+// interface SetFilterAction {
+//   type: typeof SET_FILTER;
+//   payload: Filter;
+// }
+
+// interface SetSearchTextAction {
+//   type: typeof SET_SEARCH_TEXT;
+//   payload: string;
+// }
```
choeqq, line 15: why are you removing necessary types for redux?

src/features/todos.ts
```diff
@@ -1,9 +1,44 @@
...
+export const SET_TODOS = 'todos/SET_TODOS';
 
-const todosReducer = (): Todo[] => {
-  return [];
+export const setTodos = (todos: Todo[]) => ({
+  type: SET_TODOS,
+  payload: todos,
+});
+
+export const fetchTodos = () => async (dispatch: any) => {
```
choeqq, line 12: dispatch is a specific redux function which should take some redux actions

choeqq (changes requested):

PR #740, opened 2023-12-20
result

errisi:
[DEMO LINK](https://errisi.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,4 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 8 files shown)


src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,81 @@
...
+              {currentTodo?.completed && (
+                <strong className="has-text-success">Done</strong>
+              )}
 
-            {/* For completed */}
-            <strong className="has-text-success">Done</strong>
-            {' by '}
-            <a href="mailto:Sincere@april.biz">Leanne Graham</a>
-          </p>
+              {user && (
```
yevtok, line 67: instead of using a condition and a fragment it is better to write <a href={`mailto:${user?.email}`}>{user?.name}</a>

yevtok (approved):
Good job!

src/components/TodoModal/TodoModal.tsx
BudnikOleksii: > instead of using a condition and a fragment it is better to write <a href={`mailto:${user?.email}`}>{user?.name}

No, it's not better, in this case you could have invalid `href`

src/App.tsx
```diff
@@ -1,14 +1,39 @@
...
+      .then(todosFromServer => {
+        dispatch(todosActions.setTodos(todosFromServer));
+      })
+      .catch((error) => {
+        throw error;
+      })
+      .finally(() => {
+        setIsLoading(false);
+      });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
BudnikOleksii, line 32: Don't disable eslint

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,101 @@
...
+              <th>Title</th>
+              <th aria-label="no"> </th>
+            </tr>
+          </thead>
+
+          <tbody>
+            {todos.map(todo => (
+              <tr
+                key={todo.id}
+                data-cy="todo"
```
BudnikOleksii, line 48: Created a todo component

BudnikOleksii (approved):
Good work 👍 
Approved, but left a few suggestions, so check them

PR #744, opened 2023-12-24
solution

yevtok:
[DEMO LINK](https://yevtok.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,210 +1,131 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
+import cn from 'classnames';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { getTodos } from '../../api';
+import { actions as todosActions } from '../../features/todos';
+import { actions as currentTodoActions } from '../../features/currentTodo';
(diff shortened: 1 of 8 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,131 @@
...
+
+  const filter = useAppSelector(state => state.filter);
+  const { status, query } = filter;
+
+  const filteredTodos = useMemo(() => {
+    let preparedTodos = [...todos];
+
+    preparedTodos = preparedTodos.filter(todo => {
+      switch (status) {
+        case 'all':
```
BudnikOleksii, line 38: Use enum instead of magic values

src/features/filter.ts
```diff
@@ -1,10 +1,56 @@
-export const actions = { /* put action creators here */};
+import { Status } from '../types/Status';
 
-const filterReducer = () => {
-  return {
-    query: '',
-    status: 'all',
-  };
+type SetStatusAction = {
+  type: 'filter/STATUS',
```
BudnikOleksii, line 4: I think the same, you can create and use enum

BudnikOleksii (approved):
Good work 👍 
Use enums instead of magic values.
And when you work with native redux you should always return state from the reducer, for this task it's ok, but keep in mind it

PR #747, opened 2023-12-26
solutuion

ArtemSydun:
[demo](https://artemsydun.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,4 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,14 +1,39 @@
...
+      .then(todosFromServer => {
+        dispatch(todosActions.setTodos(todosFromServer));
+      })
+      .catch((error) => {
+        throw error;
+      })
+      .finally(() => {
+        setIsLoading(false);
+      });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
polosanya, line 32: Try not to disable eslint without extra need

src/App.tsx
```diff
@@ -21,14 +46,19 @@ export const App: React.FC = () => {
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading && (
+                <Loader />
+              )}
+
+              {!isLoading && (
```
polosanya, line 53: You can use ternary operator here instead

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,17 +1,41 @@
...
+    event: React.ChangeEvent<HTMLSelectElement>,
+  ) => {
+    dispatch(filterActions.setStatus(event.target.value as Status));
+  };
+
+  const handleOnQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    dispatch(filterActions.setQuery(event.target.value));
+  };
+
+  const handleOnDelete = () => {
```
polosanya, line 20: https://javascript.plainenglish.io/handy-naming-conventions-for-event-handler-functions-props-in-react-fc1cbb791364

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,17 +1,41 @@
...
-            <option value="all">All</option>
-            <option value="active">Active</option>
-            <option value="completed">Completed</option>
+          <select
+            data-cy="statusSelect"
+            value={status}
+            onChange={handleOnStatusChange}
+          >
+            <option value={Status.All}>All</option>
+            <option value={Status.Active}>Active</option>
```
polosanya, line 37: You can use Enum for text values too

src/features/currentTodo.ts
```diff
@@ -28,8 +28,10 @@ const currentTodoReducer = (
   action: Action,
 ): State => {
   switch (action.type) {
-    // Implement all actions here
-
+    case 'currentTodo/SET':
```
polosanya, line 31: Better create Enum for action types too

polosanya (changes requested):
Good job! Let's improve your code a bit)

src/features/currentTodo.ts
ArtemSydun: can we put objects into enum?

BudnikOleksii: > can we put objects into enum?

No, and you don't need to store objects in enum, it doesn't make sense

BudnikOleksii (approved):
Good work 👍

PR #751, opened 2023-12-28
Redux solution

YaroslavYarynych:
Hi, check my work:
[DEMO LINK](https://YaroslavYarynych.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,32 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
+import { actions as todosActions } from './features/todos';
+import { getTodos } from './api';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
-import { TodoModal } from './components/TodoModal';
+import { useAppDispatch, useAppSelector } from './app/hooks';
+import { Todo } from './types/Todo';
 import { Loader } from './components/Loader';
+import { TodoModal } from './components/TodoModal';
 
 export const App: React.FC = () => {
+  const [isLoading, setIsLoading] = useState(true);
+
+  const dispatch = useAppDispatch();
+  const todos: Todo[] = useAppSelector(state => state.todos);
+  const currentTodo: Todo | null = useAppSelector(state => state.currentTodo);
+
+  useEffect(() => {
+    getTodos()
+      .then((data) => {
+        dispatch(todosActions.loadTodos(data));
+      })
+      .finally(() => setIsLoading(false));
+  }, [dispatch]);
+
   return (
     <>
       <div className="section">
@@ -21,14 +39,18 @@ export const App: React.FC = () => {
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading ? (
+                <Loader />
+              )
+                : (<TodoList todos={todos} dispatch={dispatch} currentTodo={currentTodo} />)}
             </div>
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {currentTodo && (
+        <TodoModal todo={currentTodo} dispatch={dispatch} />
+      )}
     </>
   );
 };
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,6 +1,31 @@
-import React from 'react';
+/* eslint-disable max-len */
+import React, { useEffect, useState } from 'react';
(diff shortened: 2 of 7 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,103 @@
...
-            <td className="has-text-right is-vcentered">
-              <button data-cy="selectButton" className="button" type="button">
-                <span className="icon">
-                  <i className="far fa-eye" />
-                </span>
-              </button>
-            </td>
-          </tr>
+  const filteredTodos = todos.filter(todo => searchValidation(todo)
+    && (filter.status === 'all'
```
ivaZaiets, line 24: it would be better to use ```enum``` for ```all, active, completed```

src/App.tsx
```diff
@@ -21,14 +39,18 @@ export const App: React.FC = () => {
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {isLoading ? (
+                <Loader />
+              )
+                : (<TodoList todos={todos} dispatch={dispatch} currentTodo={currentTodo} />)}
```
ivaZaiets, line 45: I think, if you use ```Redax```, you can write without props

ivaZaiets (approved):
Good job!

src/App.tsx
```diff
@@ -1,14 +1,31 @@
...
+import { useAppDispatch, useAppSelector } from './app/hooks';
+import { Todo } from './types/Todo';
 import { Loader } from './components/Loader';
+import { TodoModal } from './components/TodoModal';
 
 export const App: React.FC = () => {
+  const [isLoading, setIsLoading] = useState(true);
+
+  const dispatch = useAppDispatch();
+  const currentTodo: Todo | null = useAppSelector(state => state.currentTodo);
```
superpooperxxx, line 19: You don't need to type the `currentTodo', because if you did everything correctly in reducer file, then it knows the type.

src/App.tsx
```diff
@@ -1,14 +1,31 @@
...
+
+  useEffect(() => {
+    getTodos()
+      .then((data) => {
+        dispatch(todosActions.loadTodos(data));
+      })
+      .finally(() => setIsLoading(false));
```
superpooperxxx, line 26: Don't forget to handle errors

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,17 +1,43 @@
...
+import React, { useEffect, useState } from 'react';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { actions as filterActions } from '../../features/filter';
+import { Select } from '../../types/helper';
 
 export const TodoFilter: React.FC = () => {
+  const dispatch = useAppDispatch();
+  const filter = useAppSelector(state => state.filter);
+  const [status, setStatus] = useState(filter.status);
+  const [query, setQuery] = useState(filter.query);
```
superpooperxxx, line 11: If you defined `status' and `query` as parts of `filter reducer` then why you handle them as regular state? Create actions for setting `status' and `query` and trigger the dispatch

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,17 +1,43 @@
...
+
+  useEffect(() => {
+    dispatch(filterActions.setQuery(query));
+  }, [query, dispatch]);
```
superpooperxxx, line 15: Get rid of this

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,17 +1,43 @@
...
+
+    setQuery(event.target.value);
+  };
+
+  const handleStatusChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    const newStatus = event.target.value;
+
+    setStatus(newStatus);
+    dispatch(filterActions.setStatus(newStatus));
+  };
```
superpooperxxx, line 28: trigger dispatch with respective actions

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,17 +1,43 @@
...
       <p className="control">
         <span className="select">
-          <select data-cy="statusSelect">
-            <option value="all">All</option>
-            <option value="active">Active</option>
-            <option value="completed">Completed</option>
+          <select data-cy="statusSelect" value={status} onChange={handleStatusChange}>
+            <option value={Select.all}>All</option>
+            <option value={Select.active}>Active</option>
+            <option value={Select.completed}>Completed</option>
```
superpooperxxx, line 40: As it is an `enum` then map through it

superpooperxxx (changes requested):
Not bad, my boy🫡

PR #752, opened 2023-12-28
react_redux-list-of-todos task solution

ivaZaiets:
Hey! Check my solution :)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,13 @@
-/* eslint-disable max-len */
-import React from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
+
+import { useEffect } from 'react';
+
+import { useAppDispatch, useAppSelector } from './app/hooks';
+import { actions as todosActions } from './features/todos';
+
+import { getTodos } from './api';
+import { Todo } from './types/Todo';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
@@ -9,26 +15,43 @@ import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
 
 export const App: React.FC = () => {
+  const dispatch = useAppDispatch();
+  const todos: Todo[] = useAppSelector(state => state.todos);
+  const todo = useAppSelector(state => state.currentTodo);
+
+  useEffect(() => {
+    getTodos()
+      .then((data) => {
+        dispatch(todosActions.getTodos(data));
+      });
+  }, [dispatch]);
+
   return (
     <>
       <div className="section">
         <div className="container">
           <div className="box">
-            <h1 className="title">Todos:</h1>
-
-            <div className="block">
-              <TodoFilter />
-            </div>
-
-            <div className="block">
+            {todos.length === 0 ? (
               <Loader />
-              <TodoList />
-            </div>
+            ) : (
+              <>
+                <h1 className="title">Todos:</h1>
+
+                <div className="block">
+                  <TodoFilter />
+                </div>
+
+                <div className="block">
+                  <TodoList />
+                </div>
+              </>
+            )}
           </div>
         </div>
       </div>
 
-      <TodoModal />
+      {todo && <TodoModal />}
+
     </>
   );
 };
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,6 +1,20 @@
-import React from 'react';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { actions as filterActions } from '../../features/filter';
+
+import { Select } from '../../helpers/SelectEnum';
 
 export const TodoFilter: React.FC = () => {
+  const dispatch = useAppDispatch();
+  const data = useAppSelector(state => state.filter);
+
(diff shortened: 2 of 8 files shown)


YaroslavYarynych (approved):
An absolute coding symphony🤟

src/App.tsx
```diff
@@ -1,34 +1,57 @@
...
 export const App: React.FC = () => {
+  const dispatch = useAppDispatch();
+  const todos: Todo[] = useAppSelector(state => state.todos);
+  const todo = useAppSelector(state => state.currentTodo);
+
+  useEffect(() => {
+    getTodos()
+      .then((data) => {
+        dispatch(todosActions.getTodos(data));
+      });
```
superpooperxxx, line 26: Don't forget to handle errors

src/App.tsx
```diff
@@ -1,34 +1,57 @@
...
+  }, [dispatch]);
+
   return (
     <>
       <div className="section">
         <div className="container">
           <div className="box">
-            <h1 className="title">Todos:</h1>
+            {todos.length === 0 ? (
```
superpooperxxx, line 34: Be careful with such checks, because todos length can be 0 and it doesn't mean you need to show infinite loader. Create separate state `isLoaded` to check if data was loaded, if not - show the loader

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,24 +1,43 @@
...
+  const dispatch = useAppDispatch();
+  const data = useAppSelector(state => state.filter);
+
+  const handleFunc = (e: React.ChangeEvent<HTMLInputElement>) => {
+    return dispatch(filterActions.filterTodo(e.target.value));
+  };
+
+  const handleSelectFunc = (e: React.ChangeEvent<HTMLSelectElement>) => {
+    return dispatch(filterActions.selectTodo(e.target.value));
+  };
```
superpooperxxx, line 16: 1. Bad naming `handleFunc` and `handleSelectFunc`
2. These function should be responsible for setting `query` and `status` and it is not obvious from action creator name that they do this.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,24 +1,43 @@
...
-            <option value="all">All</option>
-            <option value="active">Active</option>
-            <option value="completed">Completed</option>
+          <select
+            data-cy="statusSelect"
+            onChange={handleSelectFunc}
+          >
+            <option value={Select.all}>All</option>
+            <option value={Select.active}>Active</option>
+            <option value={Select.completed}>Completed</option>
```
superpooperxxx, line 31: create `enum` for this and map though it

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -28,12 +47,17 @@ export const TodoFilter: React.FC = () => {
...
-            type="button"
-            className="delete"
-          />
+
+          {/* eslint-disable */}
+
+          {data.query && (
+            <button
+              data-cy="clearSearchButton"
+              onClick={() => dispatch(filterActions.removeQuery())}
```
superpooperxxx, line 56: no need it separate action creator for reseting the query, just pass `''` to action creator that handle value updating

src/features/filter.ts
```diff
@@ -1,10 +1,46 @@
...
 
-const filterReducer = () => {
-  return {
-    query: '',
-    status: 'all',
-  };
+type SelectTodosAction = { type: 'select/SELECT', payload: string };
+const selectTodo = (value: string): SelectTodosAction => ({
+  type: 'select/SELECT', payload: value,
+});
```
superpooperxxx, line 9: Within one reducer you should use one prefix. You cannot have some actions start with `select` and other with `filter`

src/App.tsx
```diff
@@ -1,34 +1,57 @@
...
 
 export const App: React.FC = () => {
+  const dispatch = useAppDispatch();
+  const todos: Todo[] = useAppSelector(state => state.todos);
+  const todo = useAppSelector(state => state.currentTodo);
+
+  useEffect(() => {
+    getTodos()
+      .then((data) => {
+        dispatch(todosActions.getTodos(data));
```
superpooperxxx, line 25: Probably `setTodos`

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,131 @@
...
+  const currentTodo = useAppSelector(state => state.currentTodo);
 
-          <tr data-cy="todo">
-            <td className="is-vcentered">8</td>
+  useEffect(() => {
+    getTodos()
+      .then((tds) => {
+        dispatch(todosActions.getTodos(tds));
+      });
+  }, [dispatch]);
```
superpooperxxx, line 24: You already loaded the data, what is the point of doing it again?

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,131 @@
...
 
-            <td className="is-vcentered">
-              <span className="icon" data-cy="iconCompleted">
-                <i className="fas fa-check" />
-              </span>
-            </td>
+      default:
+        return filteredTodos;
+    }
+  };
```
superpooperxxx, line 50: It should be one function responsible for filtering, move to outside of the component.

superpooperxxx (changes requested):
Good job, but changes are required💅🏻

PR #779, opened 2024-01-14
Develop

denzzlle171:
[DEMO LINK](https://denzzlle171.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,76 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
+import { useSelector, useDispatch } from 'react-redux';
+
+import { getTodos, getUser } from './api';//
 
 import { TodoList } from './components/TodoList';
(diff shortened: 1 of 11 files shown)


A1eeex (commented):
**fix this**

image
vs
image

src/App.tsx
```diff
@@ -1,14 +1,76 @@
...
+
+  useEffect(() => {
+    getTodos()
+      .then((fetchedTodos) => {
+        dispatch(todosActions.setTodos(fetchedTodos));
+      })
+      .finally(() => {
+        setIsLoadingTodos(false);
+      });
+  }, []);
```
A1eeex, line 43: better handle the error
add `catch {}`

src/App.tsx
```diff
@@ -1,14 +1,76 @@
...
+      dispatch(userActions.setUser(user));
+      setIsLoadingUser(false);
+    });
+  };
+
+  const filteredTodos = todos
+    .filter((todo) => {
+      switch (filterState.status) {
+        case Status.all:
+          return true;
```
A1eeex, line 59: here you can do without it,
default: returns the same result

A1eeex (changes requested):

PR #783, opened 2024-01-16
Develop

Patrykmclaren14:
[solution](https://Patrykmclaren14.github.io/react_redux-list-of-todos/)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3383,9 +3383,9 @@
       }
     },
     "@types/react-redux": {
-      "version": "7.1.32",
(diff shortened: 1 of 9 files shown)


Barosz30 (approved):
Very good code!

Tests may be mistaken by this:
{!todos.length ? <Loader />
                : <TodoList />}
                
  Try compraing it to 0 length, or something like isLoading(true/false)
  
  Code works great, it's just how tests were constructed.

mbruhler (approved):
One can say it's better to use named functions in event handlers, but I'm personally not a fan of it, I do the same. So I can't say it's a mistake (because there is not strict standard) but keep an eye on that

Approved.

PR #790, opened 2024-01-20
Solution

VKdrvtsv:
[DEMO LINK](https://VKdrvtsv.github.io/react_redux-list-of-todos/)

All redux-parts works well but I have a problem with background element in TodoModal.

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,6 +1,40 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { actions } from '../../features/filter';
+import { useAppDispatch } from '../../app/hooks';
 
 export const TodoFilter: React.FC = () => {
+  const [value, setValue] = useState('all');
+  const [searchValue, setSearchValue] = useState('');
+
+  const {
+    setStatusActive,
+    setStatusAll,
+    setStatusCompleted,
+    getQuery,
+  } = actions;
+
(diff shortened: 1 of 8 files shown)


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,14 +1,51 @@
...
+    setStatusAll,
+    setStatusCompleted,
+    getQuery,
+  } = actions;
+
+  const dispatch = useAppDispatch();
+
+  useEffect(() => {
+    switch (value) {
+      case 'active':
```
anastasiia-tilikina, line 20: use enum for status

src/features/filter.ts
```diff
@@ -1,10 +1,83 @@
-export const actions = { /* put action creators here */};
+import { Status } from '../types/Status';
 
-const filterReducer = () => {
-  return {
+type SetAllFilterAction = { type: 'filter/ALL' };
+
+const setStatusAll = (): SetAllFilterAction => ({
```
anastasiia-tilikina, line 5: you can create only one method - setStatus and use value as payload

anastasiia-tilikina (changes requested):
there is some problem with modal. I believe this is the reason of test falling
https://www.loom.com/share/74fc0f0640ad40b6b76a9276c6d66801?sid=e82869c1-7b63-4ab1-9ab4-14cab8fb50ce

VitaliyBondarenko1982 (approved):
Looks good.

PR #803, opened 2024-02-07
add react_redux-list-of-todos

Dizelkos:
[DEMO LINK](https://Dizelkos.github.io/react_redux-list-of-todos/)


src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,72 @@
...
-          </p>
+            <p className="block" data-cy="modal-user">
+              {currentTodo?.completed ? (
+                <strong className="has-text-success">Done</strong>
+              ) : (
+                <strong className="has-text-danger">Planned</strong>
+              )}
+              {' by '}
+              <a href={`mailto:${user?.email}`}>
+                {user?.name}
```
volodymyr-soltys97, line 64: Use destruction for `user`

src/features/filter.ts
```diff
@@ -1,10 +1,66 @@
...
+const filterReducer = (
+  state = {
     query: '',
     status: 'all',
-  };
+  },
+  action: Action,
+) => {
+  switch (action.type) {
+    case 'filter/ALL':
```
TarasHoliuk, line 37: it would look a bit simpler with one action `filter/STATUS` that accepts payload (status) that should be set (same as QUERY action)

TarasHoliuk (approved):
1 comment FYI

PR #812, opened 2024-02-15
add task solution

avakiel:
[DEMO LINK](https://avakiel.github.io/react_redux-list-of-todos/)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,101 @@
...
-                <span className="icon">
-                  <i className="far fa-eye" />
-                </span>
-              </button>
-            </td>
-          </tr>
+export const TodoList = () => {
+  const currentTodo = useAppSelector(state => state.currentTodo);
+  const todos = useAppSelector(state => state.todos);
+  const filter = useAppSelector(state => state.filter);
```
artem8746, line 12:
```suggestion
  const { currentTodo, todos, filter  } = useAppSelector(state => state);
```
you can use destructuring here

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,101 @@
...
+  const filtredTodos = () => {
+    switch (filter.status) {
+      case 'all':
+        return todos;
 
-            <td className="is-vcentered is-expanded">
-              <p className="has-text-danger">fugiat veniam minus</p>
-            </td>
+      case 'active':
+        return todos?.filter(todo => todo.completed === false);
```
artem8746, line 21: would be better to create enum for todo state

artem8746:
image
hide table on loading

src/components/TodoList/TodoList.tsx
```diff
@@ -1,210 +1,101 @@
...
+        return todos;
+    }
+  };
 
-          <tr data-cy="todo">
-            <td className="is-vcentered">5</td>
-            <td className="is-vcentered"> </td>
+  const queryFilter = (todoList: Todo[]) => {
+    return todoList?.filter(todo => todo.title.toLowerCase().includes(filter.query.toLowerCase().trim()));
+  };
```
artem8746, line 33: it's too long, divide it, for example create constant for normalizedQuery

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,69 @@
...
+import { useEffect, useState } from 'react';
 import { Loader } from '../Loader';
+import { User } from '../../types/User';
+import { getUser } from '../../api';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { actions } from '../../features/currentTodo';
+
+export const TodoModal = () => {
+  const [user, setUser] = useState<User>();
+  const [isLoading, setIsLoading] = useState(false);
```
artem8746, line 10: would be better to set true as initial state, to avoid microrendering

artem8746 (changes requested):
☜(ﾟヮﾟ☜)
Excellent job, just a few fixes needed

PR #815, opened 2024-02-20
task done

mykhailoyurkovskyi:
Demo - https://mykhailoyurkovskyi.github.io/react_redux-list-of-todos/

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,6 +1,13 @@
 import React from 'react';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { actions as filterActions } from '../../features/filter';
+import { Status } from '../../types/Status';
 
 export const TodoFilter: React.FC = () => {
+  const { status } = useAppSelector(state => state.filter);
+  const { query } = useAppSelector(state => state.filter);
+  const dispatch = useAppDispatch();
+
   return (
     <form
       className="field has-addons"
@@ -8,7 +15,13 @@ export const TodoFilter: React.FC = () => {
     >
       <p className="control">
         <span className="select">
-          <select data-cy="statusSelect">
+          <select
+            data-cy="statusSelect"
+            value={status}
+            onChange={e =>
+              dispatch(filterActions.setStatus(e?.target.value as Status))
+            }
+          >
             <option value="all">All</option>
             <option value="active">Active</option>
             <option value="completed">Completed</option>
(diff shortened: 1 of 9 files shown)


YuriySobakar (approved):
Good job! But one thing must be fixed. When filtering by query and have no matches you must hide the whole table and show only  this:
image

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -22,19 +35,26 @@ export const TodoFilter: React.FC = () => {
...
-          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-          <button
-            data-cy="clearSearchButton"
-            type="button"
-            className="delete"
-          />
-        </span>
+        {query && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
roman-mirzoian, line 49: Try to rewrite the code so that it is not necessary to disable the linter

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -22,19 +35,26 @@ export const TodoFilter: React.FC = () => {
...
-        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-          <button
-            data-cy="clearSearchButton"
-            type="button"
-            className="delete"
-          />
-        </span>
+        {query && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
```
roman-mirzoian, line 48: Avoid using inline styles.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,12 +1,41 @@
-/* eslint-disable */
-import React from 'react';
+/* eslint-disable jsx-a11y/control-has-associated-label */
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,12 +1,41 @@
...
+      default:
+        return todo;
+    }
+  });
+
+  const visibleTodos = filteredTodos.filter(todo => {
+    return todo.title.toLowerCase().includes(query.toLowerCase());
+  });
+
+  const badQuery = visibleTodos.length < 1;
```
roman-mirzoian, line 30:
```suggestion
  const isTodoListEmpty = visibleTodos.length < 1;
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,42 +1,72 @@
...
+              data-cy="modal-header"
+            >
+              {`Todo #${todo?.id}`}
+            </div>
 
-        <div className="modal-card-body">
-          <p className="block" data-cy="modal-title">
-            fugiat veniam minus
-          </p>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
roman-mirzoian, line 39: Try to rewrite the code so that it is not necessary to disable the linter

src/features/filter.ts
```diff
@@ -1,12 +1,78 @@
...
 };
 
-const filterReducer = () => {
-  return {
-    query: '',
-    status: 'all',
-  };
+const filterReducer = (state: State = initialState, action: Action) => {
+  switch (action.type) {
+    case 'filter/SET_STATUS':
```
roman-mirzoian, line 53: It is a good practice to put the names of actions in separate constants.

roman-mirzoian (changes requested):
Good work, but there are a few issues left.
Also fix the comment from the previous check by Yuriy.

PR #816, opened 2024-02-21
Solution

SashaViun:
[DEMO LINK](https://SashaViun.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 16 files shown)


src/App.tsx
```diff
@@ -1,14 +1,68 @@
...
+  const handleSelectTodo = useCallback(
+    (todo: Todo) => {
+      dispatch(actions.setTodo(todo));
+    },
+    [dispatch],
+  );
+
+  const handleDeselectTodo = useCallback(() => {
+    dispatch(actions.removeTodo());
+  }, [dispatch]);
```
DarkMistyRoom, line 64: you use Redux to avoid passing props. Dispatch these actions inside components where they are needed

DarkMistyRoom (changes requested):
- the demo link doesn't work

image

- move these folders from `components` to `utils` 
image

PR #818, opened 2024-02-21
react_redux-list-of-todos

serjgrm:
[DEMO LINK](https://<serjgrm>.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,16 @@
 /* eslint-disable max-len */
+
 import React from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
-
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
-import { Loader } from './components/Loader';
+// import { useAppSelector } from './app/hooks';
 
 export const App: React.FC = () => {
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,14 +1,16 @@
...
 import '@fortawesome/fontawesome-free/css/all.css';
-
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
-import { Loader } from './components/Loader';
+// import { useAppSelector } from './app/hooks';
 
 export const App: React.FC = () => {
+  // const { currentTodo } = useAppSelector(state => state.currentTodo);
```
sTorba24, line 12: remove comments

sTorba24 (changes requested):
Good work! Lets improve it 

1. Shouldn't be case-sensitive
image


2. Also event when i clear input after it elements not appear
image

DarkMistyRoom (changes requested):
- show Loader before todos are loaded

- don't use json file to get todos, use getTodos function from `api.ts`
- show the hide button for the selected todo
image

- hide this button when the search field is empty
image

- The clear button doesn't work

- green ones are completed and red ones are active
image

feel free to ask in the chat if something is unclear!

l4st1m0za (approved):
Good job, the only thing to make is input should not be case-sensitive

PR #819, opened 2024-02-22
Develop

MarekSax:
[DEMO](https://mareksax.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,30 +1,33 @@
 /* eslint-disable max-len */
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
-import { effect, signal } from '@preact/signals-react';
-import { useSignals } from '@preact/signals-react/runtime';
-import { StrictMode } from 'react';
+import { useEffect, useState } from 'react';
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { Loader } from './components/Loader';
 import { getTodos } from './api';
 import { TodoModal } from './components/TodoModal';
-import { selectedTodo, todos } from './signals';
-
-const loading = signal<boolean>(true);
-
-effect(() => {
-  getTodos()
-    .then(t => {
-      todos.value = t;
-    })
-    .then(() => {
-      loading.value = false;
-    });
-});
+import { useAppDispatch, useAppSelector } from './app/hooks';
+import { actions } from './features/todos';
 
 export const App: React.FC = () => {
-  useSignals();
+  const dispatch = useAppDispatch();
+  const currentTodo = useAppSelector(state => state.currentTodo);
+
+  const [loading, setLoading] = useState<boolean>(true);
+
+  useEffect(() => {
+    getTodos()
+      .then(t => {
+        dispatch(actions.setTodos(t));
+      })
+      .catch(() => {
+        throw new Error('Failed to fetch todos');
+      })
+      .finally(() => {
+        setLoading(false);
+      });
+  }, [dispatch]);
 
   return (
     <>
@@ -37,13 +40,11 @@ export const App: React.FC = () => {
               <TodoFilter />
             </div>
 
-            <div className="block">
-              {loading.value ? <Loader /> : <TodoList />}
-            </div>
+            <div className="block">{loading ? <Loader /> : <TodoList />}</div>
           </div>
         </div>
       </div>
-      <StrictMode>{!!selectedTodo.value && <TodoModal />}</StrictMode>
+      {currentTodo && <TodoModal />}
     </>
   );
 };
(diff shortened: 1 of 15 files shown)


src/App.tsx
```diff
@@ -1,14 +1,34 @@
...
+import { getTodos } from './api';
+import { TodoModal } from './components/TodoModal';
+import { useAppDispatch, useAppSelector } from './app/hooks';
+import { actions } from './features/todos';
 
 export const App: React.FC = () => {
+  const dispatch = useAppDispatch();
+  const currentTodo = useAppSelector(state => state.currentTodo);
+
+  const [loading, setLoading] = useState<boolean>(true);
```
yevhenii-pyl, line 17: Not sure that you actually need this loading state here. 
Working with redux, most probably, you will have something like todosLoading in your store that will be toggled through actions. 

However, to emulate this approach here, I would also add a `setLoading(true)` in `useEffect` before fetching. 
This will be more explicit approach, without having to consider a default state.

src/App.tsx
```diff
@@ -1,14 +1,34 @@
...
+
+  useEffect(() => {
+    getTodos()
+      .then(t => {
+        dispatch(actions.setTodos(t));
+      })
+      .catch(() => {
+        throw new Error('Failed to fetch todos');
```
yevhenii-pyl, line 25: That's rude :)) 

Once again, in a fully set environment you will probably have something like a loadingTotosError in state, however, we have what we have so why don't we use `useState` once again to store a possible error and adjust our UI to display it? 
I mean, user should be aware if something went wrong.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,17 +1,35 @@
-import React from 'react';
+import { Filters, KeyboardKeys } from '../../types';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { actions } from '../../features/filter';
 
 export const TodoFilter: React.FC = () => {
+  const dispatch = useAppDispatch();
+  const filter = useAppSelector(state => state.filter);
+
+  const handleInput = (e: React.ChangeEvent<HTMLInputElement>) => {
```
yevhenii-pyl, line 9: Nice, having handlers separated from a rendering part is always a good idea.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -20,21 +38,28 @@ export const TodoFilter: React.FC = () => {
...
 
-        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-          <button
-            data-cy="clearSearchButton"
-            type="button"
-            className="delete"
-          />
-        </span>
+        {!!filter.query.length && (
```
yevhenii-pyl, line 52: 👍

src/components/TodoList/TodoList.tsx
```diff
@@ -1,224 +1,101 @@
...
+              <i className="fas fa-check" />
+            </span>
+          </th>
+          <th>Title</th>
+          <th aria-label="empty header"> </th>
+        </tr>
+      </thead>
+
+      <tbody>
+        {filteredTodos.map(todo => (
```
yevhenii-pyl, line 51: Hmm, probably I would recommend creating a separate component for `<Todo />` here, just to avoid overcomplicating this component.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,42 +1,79 @@
...
+import { User } from '../../types';
 
 export const TodoModal: React.FC = () => {
+  const [modalLoading, setModalLoading] = useState<boolean>(true);
+  const [user, setUser] = useState<User | null>(null);
+
+  const currentTodo = useAppSelector(state => state.currentTodo);
+  const dispatch = useAppDispatch();
+
+  useEffect(() => {
```
yevhenii-pyl, line 15: Same advice with loader here.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,42 +1,79 @@
...
+        .then(data => {
+          setUser(data);
+        })
+        .then(() => {
+          setModalLoading(false);
+        });
+    }
+  }, [currentTodo]);
+
+  const handleCloseModal = () => {
```
yevhenii-pyl, line 27: 👍

src/features/filter.ts
```diff
@@ -1,12 +1,78 @@
+import { Filters } from '../types';
```
yevhenii-pyl, line 1: It was not required in this task, but as a general rule of thumb I'd prefer having different folder structure for the project. 
IDK why is it recommended to put this type of code into `features` folder, I believe this approach is not the best example of folder structure. 

```
// actions 
   => filters
// reducers 
  = > filters 
```

yevhenii-pyl (changes requested):
Good job! 
Let's just adjust a few things here and we're good to go.

yevhenii-pyl (approved):
Solid! 🚀

PR #825, opened 2024-02-28
Implemented todos list using redux

MrHunterBoi:
[DEMO LINK](https://MrHunterBoi.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,224 +1,110 @@
 /* eslint-disable */
 import React from 'react';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { actions } from '../../features/currentTodo';
 
 export const TodoList: React.FC = () => {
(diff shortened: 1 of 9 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,224 +1,110 @@
 /* eslint-disable */
```

src/features/currentTodo.ts
```diff
@@ -25,7 +25,11 @@ type Action = SetTodoAction | RemoveTodoAction;
 
 const currentTodoReducer = (state: State = null, action: Action): State => {
   switch (action.type) {
-    // Implement all actions here
+    case 'currentTodo/SET':
```
roman-mirzoian, line 28: It is customary to put action types as separate constants.
Update in all code.

roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.
Also, not all tests currently run successfully

PR #830, opened 2024-03-07
Develop

Anton-Kuchmasov:
DEMO: https://anton-kuchmasov.github.io/react_redux-list-of-todos/


src/App.tsx
```diff
@@ -1,14 +1,30 @@
...
+  const chosenTodo = useAppSelector(state => state.currentTodo);
+
+  const [isLoading, setIsLoading] = useState(false);
+
+  useEffect(() => {
+    getTodos()
+      .then(todos => {
+        dispatch(actions.setTodos(todos));
+      })
+      .finally(() => setIsLoading(true));
```
roman-mirzoian, line 25: In asynchronous requests, it is always necessary to assume that the request will end with an error

src/components/TodoList/TodoList.tsx
```diff
@@ -1,224 +1,84 @@
-/* eslint-disable */
-import React from 'react';
+/* eslint-disable jsx-a11y/control-has-associated-label */
```

roman-mirzoian (changes requested):
If the search does not find any records, there should be an inscription "There are no todos matching current filter criteria"

PR #837, opened 2024-03-13
reduxTodoList

edkido:
[DEMO LINK](https://edkido.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,6 +1,25 @@
 import React from 'react';
+import { useDispatch } from 'react-redux';
+import { useAppSelector } from '../../app/hooks';
+import { actions } from '../../features/filter';
+import { Status } from '../../types/Status';
 
 export const TodoFilter: React.FC = () => {
+  const dispatch = useDispatch();
+  const { query, status } = useAppSelector(state => state.filter);
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    dispatch(actions.setQuery(event.target.value));
+  };
+
+  const handleStatusChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    dispatch(actions.setStatus(event.target.value as Status));
+  };
+
+  const clear = () => {
+    dispatch(actions.clearQuery());
+  };
+
   return (
     <form
       className="field has-addons"
@@ -8,7 +27,11 @@ export const TodoFilter: React.FC = () => {
     >
       <p className="control">
         <span className="select">
-          <select data-cy="statusSelect">
+          <select
+            data-cy="statusSelect"
+            onChange={e => handleStatusChange(e)}
+            value={status}
+          >
             <option value="all">All</option>
             <option value="active">Active</option>
             <option value="completed">Completed</option>
@@ -22,19 +45,24 @@ export const TodoFilter: React.FC = () => {
           type="text"
           className="input"
           placeholder="Search..."
+          value={query}
+          onChange={e => handleQueryChange(e)}
         />
         <span className="icon is-left">
           <i className="fas fa-magnifying-glass" />
         </span>
 
-        <span className="icon is-right" style={{ pointerEvents: 'all' }}>
-          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-          <button
-            data-cy="clearSearchButton"
-            type="button"
-            className="delete"
-          />
-        </span>
+        {!!query.length && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+            <button
+              data-cy="clearSearchButton"
+              type="button"
+              className="delete"
+              onClick={clear}
+            />
+          </span>
+        )}
       </p>
     </form>
   );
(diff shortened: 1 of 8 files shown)


IvanFesenko (changes requested):
Looks good. It should be fixed, only selected should have such icon
2024-03-13_16h17_45

src/components/TodoList/TodoList.tsx
```diff
@@ -12,25 +12,27 @@ export const TodoList: React.FC = () => {
...
-          return todos;
-        case 'active':
-          return !todo.completed;
-        case 'completed':
-          return todo.completed;
-        default:
-          return todo;
-      }
-    });
+    const visibleTodos = [...todos];
```
TarasHoliuk, line 15: again, no need to do a copy:
>Array.fliter do not mutate original array, making copy is redundant

src/components/TodoList/TodoList.tsx
```diff
@@ -12,25 +12,27 @@ export const TodoList: React.FC = () => {
...
+      .filter(todo => {
+        switch (status) {
+          case 'active':
+            return !todo.completed;
+
+          case 'completed':
+            return todo.completed;
+
+          default:
+            return todos;
```
TarasHoliuk, line 30: todos here will be inexplictly transformed into boolean (since it's array `!!todos` always will be `true`), but it's better to explicitly write the correct boolean value: `return true;`

> Array.fliter callback should return boolean - you have case when you return an array and an object

src/components/TodoList/TodoList.tsx
```diff
@@ -1,224 +1,109 @@
...
+  const dispatch = useDispatch();
+  const { query, status } = useAppSelector(state => state.filter);
+  const currentTodo = useAppSelector(state => state.currentTodo);
+  const todos = useAppSelector(state => state.todos);
+
+  const filteredTodos = () => {
+    const visibleTodos = [...todos];
+
+    return visibleTodos
+      .filter(todo =>
```
TarasHoliuk, line 18: > Also no need to filter in case query is empty
with such approach, you anyway filter todos, even if there is no query

Do something like that to avoid it

```
let filteredTodos = todos;

if (query.trim()) {
  filteredTodos = filteredTodos.filter(// filter by query)
}
```
Also, you may use similar approach to filter by status:
>Also no need to filter in case status all

TarasHoliuk (changes requested):

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,14 +1,37 @@
...
+import { useDispatch } from 'react-redux';
+import { useAppSelector } from '../../app/hooks';
+import { actions } from '../../features/filter';
+import { Status } from '../../types/Status';
 
 export const TodoFilter: React.FC = () => {
+  const dispatch = useDispatch();
+  const { query, status } = useAppSelector(state => state.filter);
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
GUSILLUS, line 11: You can import every event type from react library like this `import { ChangeEvent } from 'react';`

It will make your code shorter and improve readability

```suggestion
  const handleQueryChange = (event: ChangeEvent<HTMLInputElement>) => {
```

src/components/TodoList/TodoList.tsx
```diff
@@ -1,224 +1,111 @@
...
+          return !todo.completed;
+
+        case 'completed':
+          return todo.completed;
+
+        default:
+          return true;
+      }
+    });
+  }
```
GUSILLUS, line 35: Consider moving this to a separate function

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,42 +1,73 @@
...
+  const [user, setUser] = useState<User | null>(null);
+
+  const dispatch = useDispatch();
+  const currentTodo = useAppSelector(state => state.currentTodo);
+
+  useEffect(() => {
+    if (currentTodo) {
+      getUser(currentTodo.userId)
+        .then(setUser)
+        .finally(() => setIsLoading(true));
```
GUSILLUS, line 20: Consider writing true like the initial value and false into the finally block

Because right now when you get a user you have state isLoading true and I think that you have endless loading

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,42 +1,73 @@
...
+            />
+          </header>
 
-        <div className="modal-card-body">
-          <p className="block" data-cy="modal-title">
-            fugiat veniam minus
-          </p>
+          <div className="modal-card-body">
+            <p className="block" data-cy="modal-title">
+              {currentTodo?.title}
```
GUSILLUS, line 55: Consider using destruction assignment with currentTodo

GUSILLUS (changes requested):
GJ! Just fix the points that I mention

src/components/TodoModal/TodoModal.tsx
edkido: no, i don't. There is a test for this case and my loader turning off after promise is resolved

src/components/TodoModal/TodoModal.tsx
edkido: i can't, because currentTodo is typeof State

Andriy-Shymkiv (approved):
good job!

PR #845, opened 2024-03-21
add task solution

valeriaoleynik:
[DEMO LINK](https://valeriaoleynik.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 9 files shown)


src/features/currentTodo.ts
```diff
@@ -25,7 +25,11 @@ type Action = SetTodoAction | RemoveTodoAction;
 
 const currentTodoReducer = (state: State = null, action: Action): State => {
   switch (action.type) {
-    // Implement all actions here
+    case 'currentTodo/SET':
```
IvanFesenko, line 28: Pay attention
2024-03-21_16h39_03

src/features/filter.ts
```diff
@@ -1,12 +1,51 @@
...
-  return {
-    query: '',
-    status: 'all',
-  };
+const initialState: State = {
+  query: '',
+  status: 'all',
+};
+
+const filterReducer = (state: State = initialState, action: Action) => {
```
IvanFesenko, line 35: Same here

src/features/todos.ts
```diff
@@ -1,9 +1,24 @@
...
+type Action = SetTodosAction;
+
+const setTodos = (todos: Todo[]): SetTodosAction => ({
+  type: 'todos/SET',
+  payload: todos,
+});
+
+export const actions = { setTodos };
+
+const todosReducer = (todos: Todo[] = [], action: Action): Todo[] => {
```
IvanFesenko, line 14: And here

IvanFesenko (changes requested):
Demo link not works
2024-03-21_16h37_51

src/App.tsx
```diff
@@ -1,14 +1,35 @@
...
+  useEffect(() => {
+    setLoading(true);
+
+    setTimeout(() => {
+      getTodos()
+        .then(data => dispatch(todosActions.setTodos(data)))
+        .catch(() => setErrorMessage('There are no todos'))
+        .finally(() => setLoading(false));
+    }, 1000);
+  }, []);
```
TarasHoliuk, line 31:
```suggestion
  }, [dispatch]);
```
it's better to just set it - FYI
image

dispatch is also memorized under the hood (same for setState functions, etc.) and never change, so the useEffect will also work only once.

TarasHoliuk (approved):

PR #847, opened 2024-03-24
React_Redux-list-of-todos

vintwp:
https://vintwp.github.io/react_redux-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable react-hooks/exhaustive-deps */
 import React, { useEffect, useState } from 'react';
 import { getTodos } from './api';
 import 'bulma/css/bulma.css';
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,58 @@
-/* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { getTodos } from './api';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
-import { TodoModal } from './components/TodoModal';
+// import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { useAppDispatch, useAppSelector } from './app/hooks';
+import { actions as todosActions } from './features/todos';
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import { TodoModal } from './components/TodoModal';
 
 export const App: React.FC = () => {
+  const [loading, setLoading] = useState<boolean>(false);
+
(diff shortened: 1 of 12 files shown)


src/features/filter.ts
```diff
@@ -1,12 +1,46 @@
...
-    status: 'all',
-  };
+const initialFilter: Filter = { status: 'all', query: '' };
+
+const filterReducer = (filter: Filter = initialFilter, action: Action) => {
+  switch (action.type) {
+    case 'active':
+    case 'completed':
+    case 'all':
+      return {
```
etojeDenys, line 35:
```suggestion
    case 'CHANGE_FILTER_VALUE':
      return {
```
why do you need three the same actions

src/App.tsx
```diff
@@ -1,14 +1,59 @@
...
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
-import { TodoModal } from './components/TodoModal';
+// import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { Todo } from './types/Todo';
+import { useAppDispatch, useAppSelector } from './app/hooks';
+import { actions as todosActions } from './features/todos';
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
etojeDenys, line 14: looks redundant

src/App.tsx
```diff
@@ -1,14 +1,59 @@
...
+/* eslint-disable react-hooks/exhaustive-deps */
+import React, { useEffect, useState } from 'react';
+import { getTodos } from './api';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
 import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
-import { TodoModal } from './components/TodoModal';
+// import { TodoModal } from './components/TodoModal';
```
etojeDenys, line 9: .

etojeDenys (changes requested):

src/features/filter.ts
vintwp: @etojeDenys Is it correct to use value (which is not in type Status) in case conditions. In my variant, there is no duplicated code and looks more clearer code? Thanks!

src/components/TodoList/TodoList.tsx
```diff
@@ -1,224 +1,46 @@
...
-                  <i className="far fa-eye" />
-                </span>
-              </button>
-            </td>
-          </tr>
-        </tbody>
-      </table>
+  return (
+    <>
+      {(status || query) && todos.length === 0 && (
```
sTorba24, line 15:
```suggestion
      {(status || query) && !todos.length && (
```

src/features/filter.ts
sTorba24: you can change type description, but no need to use filter types as a action types. action is only 1 - change filter value

then u use values from `filter`

sTorba24 (approved):
GJ 🚀

PR #862, opened 2024-04-02
Solutiion

andrii-rykal:
[DEMO LINK](https://andrii-rykal.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,14 +1,35 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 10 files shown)


src/App.tsx
```diff
@@ -1,14 +1,35 @@
...
+
+  useEffect(() => {
+    setLoading(true);
+
+    getTodos()
+      .then(todosFromServer => {
+        dispatch(todosActions.setTodos(todosFromServer));
+      })
+      .catch(error => {
+        throw new Error('Error fetching todos', error);
```
TarasHoliuk, line 26: do not throw an error here - you handled it already (or should handle it - change some state value to show the error message, etc.), so no need to throw it again. It will lead to a page crash (because you don't try to catch the error thrown here somewhere else)

FYI: take into account in the future

TarasHoliuk (approved):
Well done 👍

PR #867, opened 2024-04-06
Solved

DenisGurskiy:
https://DenisGurskiy.github.io/react_redux-list-of-todos

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,224 +1,117 @@
 /* eslint-disable */
-import React from 'react';
+import React, { useMemo } from 'react';
+import { Todo } from '../../types/Todo';
+import classNames from 'classnames';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { actions as currentTodoActions } from '../../features/currentTodo';
 
 export const TodoList: React.FC = () => {
+  const dispatch = useAppDispatch();
+
+  const todos = useAppSelector(state => state.todos);
+  const filter = useAppSelector(state => state.filter);
+  const currentTodo = useAppSelector(state => state.currentTodo);
+
+  const setTodo = (todo: Todo) => dispatch(currentTodoActions.setTodo(todo));
+
+  const filteringTodos = useMemo(() => {
+    let filterTodos = [...todos];
+
+    if (filter.query) {
+      filterTodos = filterTodos.filter(todo => {
+        return todo.title.toLowerCase().includes(filter.query.toLowerCase());
+      });
+    }
+
+    return filterTodos.filter(todo => {
+      if (filter.status === 'all') {
+        return true;
+      }
+
+      if (filter.status === 'active') {
+        return !todo.completed;
+      }
+
+      return todo.completed;
+    });
+  }, [filter, todos]);
+
+  const handleSelect = (todo: Todo) => {
+    setTodo(todo);
+  };
+
   return (
     <>
-      <p className="notification is-warning">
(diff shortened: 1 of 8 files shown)


src/features/filter.ts
```diff
@@ -1,12 +1,48 @@
...
-  /* put action creators here */
+  setFilter: (filter: Status, query: string): Action => ({
+    type: filter,
+    payload: query,
+  }),
+};
+
+const initialState = {
+  query: '',
+  status: 'all',
```
TarasHoliuk, line 14: We have a few possible statuses - let's add enum for it

src/features/todos.ts
```diff
@@ -1,9 +1,23 @@
...
+  payload: todo,
+});
+
+export const actions = { addTodos };
+
+// eslint-disable-next-line @typescript-eslint/default-param-last
+const todosReducer = (todos: Todo[] = [], action: AddAction): Todo[] => {
+  switch (action.type) {
+    case 'todos/ADD':
+      return [...todos, action.payload];
```
TarasHoliuk, line 16: since in this case, you don't add a single todo actually it would be more effective to add all todos in one go. Pass the array of todos as payload and add them all
```suggestion
      return [...todos, ...action.payload];
```
P.S. change the action name accordingly

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,42 +1,78 @@
...
 
-          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-          <button type="button" className="delete" data-cy="modal-close" />
-        </header>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+            <button
+              type="button"
+              className="delete"
+              data-cy="modal-close"
+              onClick={handleClose}
```
TarasHoliuk, line 54:
```suggestion
              onClick={removeTodo}
```
`handleClose` function wrapper is redundant

src/components/TodoList/TodoList.tsx
```diff
@@ -1,224 +1,117 @@
...
+                  >
+                    {todo.title}
+                  </p>
+                </td>
+                <td className="has-text-right is-vcentered">
+                  <button
+                    data-cy="selectButton"
+                    className="button"
+                    type="button"
+                    onClick={() => handleSelect(todo)}
```
TarasHoliuk, line 98:
```suggestion
                    onClick={() => setTodo(todo)}
```
same - handleSelect wrapper is redundant

src/components/TodoList/TodoList.tsx
```diff
@@ -1,224 +1,117 @@
...
-            <td className="has-text-right is-vcentered">
-              <button data-cy="selectButton" className="button" type="button">
-                <span className="icon">
-                  <i className="far fa-eye" />
-                </span>
-              </button>
-            </td>
-          </tr>
+  const filteringTodos = useMemo(() => {
+    let filterTodos = [...todos];
```
TarasHoliuk, line 18: you anyway will call `filter` below (which will create a new array), so making copy here is redundant
```suggestion
    let filteredTodos = todos;
```
**filtered**. filterTodos sounds like a function name

TarasHoliuk (changes requested):
Great job 👍 
Added a few comments to improve your code a bit 🙂

DenisGurskiy:
Thanks a lot. Your comments were very helpful.

TarasHoliuk (approved):
So fast response 🔥 
Nicely done - approved ✔️

PR #874, opened 2024-04-10
added redux list of todos

MartaChobaniuk:
[DEMO LINK](https://martachobaniuk.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,7 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { useDispatch } from 'react-redux';
+
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,14 +1,41 @@
...
+  const [loading, setLoading] = useState(false);
+
+  useEffect(() => {
+    setLoading(true);
+    getTodos()
+      .then(todosFromServer => {
+        dispatch(actions.setTodos(todosFromServer));
+      })
+      .catch(error => {
+        throw error;
```
TarasHoliuk, line 30: do not throw an error here - you handled it already, so no need to throw it again. It will lead to a page crash (because you don't try to catch the error thrown here somewhere else)

You may create a separate state and change it when catching an error to show a message, etc.

TarasHoliuk (approved):
Well done - approved ✔️ 
Added 1 comment just FYI

PR #880, opened 2024-04-17
Solution

antonina-klishch:
[DEMO LINK](https://antonina-klishch.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoModal/TodoModal.tsx
+++ b/src/components/TodoModal/TodoModal.tsx
@@ -1,42 +1,63 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { actions as actionsCurrentTodo } from '../../features/currentTodo';
+import { getUser } from '../../api';
 import { Loader } from '../Loader';
+import { User } from '../../types/User';
 
 export const TodoModal: React.FC = () => {
+  const { currentTodo, loading } = useAppSelector(state => state.currentTodo);
+  const [user, setUser] = useState<User | null>(null);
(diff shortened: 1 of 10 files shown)


src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,42 +1,63 @@
...
 
 export const TodoModal: React.FC = () => {
+  const { currentTodo, loading } = useAppSelector(state => state.currentTodo);
+  const [user, setUser] = useState<User | null>(null);
+  const dispatch = useAppDispatch();
+
+  useEffect(() => {
+    dispatch(actionsCurrentTodo.setLoadig(true));
+
+    getUser(currentTodo?.userId as number)
```
andriy-shymkiv, line 16: but `currentTodo?.userId` can be undefined here so need to handle this case:
```ts
const userId = currentTodo?.userId;
if (!userId) {
  throw new Error('userId is undefined')
}
getUser(userId) // no need to cast

```

andriy-shymkiv (changes requested):

src/features/currentTodo.ts
```diff
@@ -1,32 +1,72 @@
...
+};
+
+const initialState: State = {
+  currentTodo: null,
+  loading: false,
+  error: '',
+};
+
+const currentTodoReducer = (
+  state: State = initialState,
```
TarasHoliuk, line 58: That's the case when we should ignore lint error. In such cases it's better to disable the rule for this file:
image

FYI

TarasHoliuk (approved):
Approved ✔️ 
Added 1 comment just FYI

PR #899, opened 2024-05-03
Develop

zozulia-developer:
[DEMO LINK](https://zozulia-developer.github.io/react_redux-list-of-todos/)


lerastarynets (changes requested):
Such a strange behaviour, maybe consider placing loader inside modal, it definitely shouldn't appear together with content
Screenshot at May 03 13-07-20

lerastarynets (approved):
You're on the right track!

PR #903, opened 2024-05-09
Develop

ArtemZakhar:
Hello)

I am having a very unproductive week(
So I decided not to make it beauty but just pass the tests to finish faster)

[DEMO LINK](https://ArtemZakhar.github.io/react_redux-list-of-todos/)

Have a nice day)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -22,6 +22,7 @@ export const App: React.FC = () => {
         dispatch({ type: 'todos/ADDALL', payload: response });
       })
       .finally(() => setLoading(false));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
   return (
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable max-len */
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 
@@ -7,8 +7,23 @@ import { TodoList } from './components/TodoList';
 import { TodoFilter } from './components/TodoFilter';
 import { TodoModal } from './components/TodoModal';
 import { Loader } from './components/Loader';
+import { getTodos } from './api';
+import { useAppDispatch } from './app/hooks';
 
 export const App: React.FC = () => {
+  const dispatch = useAppDispatch();
+
+  const [loading, setLoading] = useState(false);
+
+  useEffect(() => {
+    setLoading(true);
+    getTodos()
+      .then(response => {
+        dispatch({ type: 'todos/ADDALL', payload: response });
+      })
+      .finally(() => setLoading(false));
+  }, []);
+
   return (
     <>
       <div className="section">
@@ -21,8 +36,8 @@ export const App: React.FC = () => {
             </div>
 
             <div className="block">
-              <Loader />
-              <TodoList />
+              {loading && <Loader />}
+              <TodoList loading={loading} />
             </div>
           </div>
         </div>
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,6 +1,11 @@
 import React from 'react';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { Status } from '../../types/Status';
 
 export const TodoFilter: React.FC = () => {
+  const dispatch = useAppDispatch();
+  const { query } = useAppSelector(state => state.filter);
+
   return (
     <form
       className="field has-addons"
@@ -8,7 +13,15 @@ export const TodoFilter: React.FC = () => {
     >
       <p className="control">
         <span className="select">
-          <select data-cy="statusSelect">
+          <select
+            data-cy="statusSelect"
+            onChange={event =>
+              dispatch({
+                type: 'filters/CHANGE',
+                payload: event.target.value as Status,
+              })
+            }
+          >
             <option value="all">All</option>
             <option value="active">Active</option>
             <option value="completed">Completed</option>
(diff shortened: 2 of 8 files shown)


src/App.tsx
```diff
@@ -1,14 +1,30 @@
...
 export const App: React.FC = () => {
+  const dispatch = useAppDispatch();
+
+  const [loading, setLoading] = useState(false);
+
+  useEffect(() => {
+    setLoading(true);
+    getTodos()
+      .then(response => {
+        dispatch({ type: 'todos/ADDALL', payload: response });
```
DanilWeda, line 22: A better way to use the named function for dispatching. ( not critical in this case)

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -21,20 +34,27 @@ export const TodoFilter: React.FC = () => {
           data-cy="searchInput"
           type="text"
           className="input"
+          value={query}
           placeholder="Search..."
+          onChange={event =>
+            dispatch({ type: 'filters/QUERY', payload: event.target.value })
+          }
```
DanilWeda, line 41: In this case, you need a handler. It will be more readable.

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -21,20 +34,27 @@ export const TodoFilter: React.FC = () => {
...
-          />
-        </span>
+        {!!query.length && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+            <button
+              data-cy="clearSearchButton"
+              type="button"
+              className="delete"
+              onClick={() => dispatch({ type: 'filters/QUERY', payload: '' })}
```
DanilWeda, line 54: In this case, you need a handler. It will be more readable.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,12 +1,27 @@
...
+
+  const filteredTodos = getSortedTodos(todos, { status, query });
 
-export const TodoList: React.FC = () => {
   return (
     <>
-      <p className="notification is-warning">
-        There are no todos matching current filter criteria
-      </p>
+      {!loading && !todos.length && (
```
DanilWeda, line 20:
```suggestion
      {(!loading && !todos.length) && (
```
If you don't use constant for several conditions try to use brackets for it. 
In the future, a person can simply add an additional condition based on a previous condition. ( not critical)

src/components/TodoList/TodoList.tsx
```diff
@@ -25,7 +40,52 @@ export const TodoList: React.FC = () => {
...
         <tbody>
-          <tr data-cy="todo">
+          {filteredTodos.map(todo => (
+            <tr
+              data-cy="todo"
+              key={todo.id}
+              {...(currentTodo &&
+                currentTodo.id === todo.id && {
+                  className: 'has-background-info-light',
+                })}
```
DanilWeda, line 50: Try to think how we can not use spread in attributes ( not critical)

src/components/TodoList/TodoList.tsx
```diff
@@ -25,7 +40,52 @@ export const TodoList: React.FC = () => {
...
+              {/* todo: change eye styles */}
+
+              <td className="has-text-right is-vcentered">
+                <button
+                  data-cy="selectButton"
+                  className="button"
+                  type="button"
+                  onClick={() =>
+                    dispatch({ type: 'currentTodo/SET', payload: todo })
+                  }
```
DanilWeda, line 77: In this case, you need a handler. It will be more readable.

src/components/TodoFilter/TodoFilter.tsx
DanilWeda, line 1: Also, tip - the best way add all actions type in constant end use it in all cases for best support code in the future and hot swapping if we need it.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,42 +1,69 @@
...
 
-          <p className="block" data-cy="modal-user">
-            {/* For not completed */}
-            <strong className="has-text-danger">Planned</strong>
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+              <button
+                type="button"
+                className="delete"
+                data-cy="modal-close"
+                onClick={() => dispatch({ type: 'currentTodo/REMOVE' })}
```
DanilWeda, line 44: In this case, you need a handler. It will be more readable. ( also, the best way to add payload to avoid side effects. For example: when you try to call action.payload)

src/features/currentTodo.ts
```diff
@@ -23,9 +23,16 @@ export const actions = { setTodo, removeTodo };
 type State = Todo | null;
 type Action = SetTodoAction | RemoveTodoAction;
 
+// eslint-disable-next-line @typescript-eslint/default-param-last
 const currentTodoReducer = (state: State = null, action: Action): State => {
   switch (action.type) {
-    // Implement all actions here
+    case 'currentTodo/SET': {
```
DanilWeda, line 29: for this case, you can also use a constant from one place with constants ( for all cases)

src/features/filter.ts
```diff
@@ -1,12 +1,57 @@
...
+});
+
 export const actions = {
-  /* put action creators here */
+  changeFilters,
+  queryFilter,
+};
+
+type State = {
+  query: string | '';
```
DanilWeda, line 23: why need `''`? it's also string.

src/components/TodoModal/TodoModal.tsx
ArtemZakhar: Please clarify what do you mean?)

src/components/TodoList/TodoList.tsx
ArtemZakhar: I am not sure I understand what I have to do?

You want me to make it like this: className: `${currentTodo && currentTodo.id === todo.id && 'has-background-info-light'}`?

src/features/filter.ts
ArtemZakhar: have no idea))

src/components/TodoFilter/TodoFilter.tsx
ArtemZakhar: I have problems with such an approach. I tried to make enum, but store was unhappy.
If I make a constant - I receive a broken Component(
image

ArtemZakhar: such error in case of enum
image

ArtemZakhar: now everything works fine)

src/components/TodoList/TodoList.tsx
alexander-ignatow: Spreading parameters is not the best idea here - as you have only one parameter (I avoid using it at all)

As long as you only have the `className` to set, you can simply use `classnames`

```
    className = {classnames({'has-background-info-light' : currentTodo?.id === todo.id)}
```

Notice, `?.` also performed `undefined`/`null` check

alexander-ignatow (approved):
Good job! 👍

PR #913, opened 2024-05-28
add solving

ordila:
[DEMO LINK](https://ordila.github.io/react_redux-list-of-todos/)


src/App.tsx
```diff
@@ -1,34 +1,59 @@
...
+        if (err instanceof Error) {
+          window.console.log(err.message);
+        }
+      } finally {
+        setisLoading(false);
+      }
+    };
+
+    getTodo();
+  }, []);
```
DanilWeda, line 36: You can disable eslint in this case.

src/helpers/getFilteredTodo.ts
```diff
@@ -0,0 +1,31 @@
...
+interface IQuery {
+  status: Status;
+  query: string;
+}
+
+export const getFilteredTodos = (initialTodos: Todo[], query: IQuery) => {
+  const { status, query: searchQuery } = query;
+
+  // Filter based on status
+  let filteredTodos = initialTodos;
```
DanilWeda, line 13: You don't need let it in this case. You can use const and then return operations with the array.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,42 +1,109 @@
...
+              <a href={`mailto:${user?.email}`}>{user?.name}</a>
+            </p>
+          </div>
         </div>
-      </div>
-    </div>
+      )}
+    </div>,
+
+    document.getElementById('modal-root')!,
```
DanilWeda, line 107: You can remove linter in this case

src/components/TodoItem/TodoItem.tsx
```diff
@@ -0,0 +1,58 @@
+import React, { useState } from 'react';
+import { Todo } from '../../types/Todo';
+import { TodoModal } from '../TodoModal';
+import { useDispatch } from 'react-redux';
+import { actions as currentTodoActions } from '../../redux/currentTodo/currentTodo';
```
DanilWeda, line 5: You can fix this lint error.

PR #920, opened 2024-06-07
Solution

Aksahir:
[DEMO LINK](https://Aksahir.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoModal/TodoModal.tsx
+++ b/src/components/TodoModal/TodoModal.tsx
@@ -1,42 +1,73 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { Loader } from '../Loader';
+import { getUser } from '../../api';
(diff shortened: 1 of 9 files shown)


src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,42 +1,73 @@
...
+    getUser(currentTodo.userId).then(setUser);
+  }
+
+  useEffect(() => {
+    setTimeout(() => {
+      if (currentTodo) {
+        setLoading(false);
+      }
+    }, 1000);
+  }, [currentTodo]);
```
etojeDenys, line 27: you need to fetch the user in useEffect and after that call `setLoading(false);`

etojeDenys (changes requested):

anastasiiavorobiova (approved):
Great work!

PR #926, opened 2024-06-12
develop

isaeev:
[DEMO LINK](https://isaeev.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,68 @@
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import { Loader, TodoFilter, TodoList, TodoModal } from './components';
+import { useEffect, useState } from 'react';
+import { getTodos } from './api';
+import { useAppDispatch, useAppSelector } from './app/hooks';
+import { todosSlice } from './features/todos';
+import { Todo } from './types/Todo';
 
-export const App = () => (
-  <>
-    <div className="section">
-      <div className="container">
-        <div className="box">
-          <h1 className="title">Todos:</h1>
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,26 +1,68 @@
...
+
+  useEffect(() => {
+    setIsLoading(true);
+    getTodos()
+      .then((data: Todo[]) => dispatch(todosSlice.actions.setTodos(data)))
+      .finally(() => setIsLoading(false));
+  }, [dispatch]);
+
+  const preparedTodos = () => {
+    let sortTodos = [];
```
TarasHoliuk, line 27:
```suggestion
    let sortedTodos = [];
```

src/App.tsx
```diff
@@ -1,26 +1,68 @@
...
+  const dispatch = useAppDispatch();
+
+  useEffect(() => {
+    setIsLoading(true);
+    getTodos()
+      .then((data: Todo[]) => dispatch(todosSlice.actions.setTodos(data)))
+      .finally(() => setIsLoading(false));
+  }, [dispatch]);
+
+  const preparedTodos = () => {
```
TarasHoliuk, line 26: this is not a good name for a function - preparedTodos sounds like it's already an array. But let's keep the name and change logic: wrap it with `useMemo` and add correct dependencies array as second argument

TarasHoliuk (changes requested):
Looks good 👍 
Added 2 small comments

Also, try to fix tests (did they run successfully locally? - if yes at least write it in the PR)

volodymyr-soltys97 (approved):
Great work 🔥

PR #934, opened 2024-06-26
Develop

ivan-gritsenko:
[DEMO LINK](https://ivan-gritsenko.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,224 +1,119 @@
 /* eslint-disable */
-import React from 'react';
+import React, { useEffect } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import { getTodos } from '../../api';
+import { todosSlice } from '../../features/todos';
+import cn from 'classnames';
+import { currentTodoSlice } from '../../features/currentTodo';
+
+function filterTodos(todos: any, query: string, status: string) {
+  let filteredTodos = todos;
+  if (query) {
+    filteredTodos = todos.filter((todo: any) =>
+      todo.title.toLowerCase().includes(query.toLowerCase()),
+    );
+  }
+  if (status !== 'all') {
+    filteredTodos = filteredTodos.filter(
+      (todo: any) => todo.completed === (status === 'completed'),
+    );
+  }
+  return filteredTodos;
+}
 
 export const TodoList: React.FC = () => {
+  const { fetchTodos } = todosSlice.actions;
+  const { setCurrentTodo } = currentTodoSlice.actions;
+  const dispatch = useDispatch();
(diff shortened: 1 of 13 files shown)


src/types/Todo.ts
```diff
@@ -1,4 +1,5 @@
 export interface Todo {
+  currentTodoReducer: any;
```
DanilWeda, line 2: currentReducer inside Todo? 
I think it`s can simplify

src/features/todos.ts
```diff
@@ -1,8 +1,22 @@
...
 
-export const todosSlice = createSlice({
+// interface TodoState {
+//   todos: Todo[];
+// }
+
+// const initialState: TodoState = {
+
+// };
+
```
DanilWeda, line 11: commented part

src/features/currentTodo.ts
```diff
@@ -1,10 +1,16 @@
...
 
-export const currentTodoSlice = createSlice({
+export const currentTodoSlice: Slice<Todo | null> = createSlice({
   name: 'currentTodo',
   initialState,
-  reducers: {},
+  reducers: {
+    setCurrentTodo(_, action: PayloadAction<Todo | null>) {
+      return action.payload;
+    },
```
DanilWeda, line 12: In my view, strange thing. 
The reducer needs to change state. 
But in this case, we do not do that. Why do we need them?

src/components/TodoList/TodoList.tsx
```diff
@@ -1,224 +1,119 @@
...
+      todo.title.toLowerCase().includes(query.toLowerCase()),
+    );
+  }
+  if (status !== 'all') {
+    filteredTodos = filteredTodos.filter(
+      (todo: any) => todo.completed === (status === 'completed'),
+    );
+  }
+  return filteredTodos;
+}
```
DanilWeda, line 22: You can move this to another file. 
Also, you need to write an enum for all statuses.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,224 +1,119 @@
...
 
 export const TodoList: React.FC = () => {
+  const { fetchTodos } = todosSlice.actions;
+  const { setCurrentTodo } = currentTodoSlice.actions;
+  const dispatch = useDispatch();
+  const todos = useSelector((state: any) => state.todosReducer);
+  const currentTodo = useSelector((state: any) => state.currentTodoReducer);
+  const { query, status } = useSelector((state: any) => state.filterReducer);
+  let visibleTodos = filterTodos(todos, query, status);
```
DanilWeda, line 31: You can use const.

DanilWeda (changes requested):
Good job. 
Please, check some comments that I left for your solution.
Many thanks!)

PR #960, opened 2024-07-22
React redux list of todos

Denys209818:
[DEMO LINK](https://denys209818.github.io/react_redux-list-of-todos)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,73 @@
+/* eslint-disable react-hooks/exhaustive-deps */
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import { Loader, TodoFilter, TodoList, TodoModal } from './components';
+import { useAppDispatch, useAppSelector } from './app/hooks';
+import { useEffect, useState } from 'react';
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,26 +1,73 @@
...
+          }
+
+          case StatusType.COMPLETED: {
+            queried = queried.filter(todo => todo.completed);
+
+            break;
+          }
+        }
+
+        dispatch(setTodos(queried));
```
SanyaBratashchuk, line 44: I think it is better to store all todos in state and do filtering on client side if filtering is not supported on BE side

SanyaBratashchuk (changes requested):

src/App.tsx
Denys209818: It was pointed out in the task condition

Denys209818: After all in this case Loader isn't appropriate because react makes calculations instantly. I agree with you that it's better in general because it requires resources to ask data from server

Denys209818: I have added a var turn on Loader if data loads

SanyaBratashchuk (approved):

PR #974, opened 2024-08-08
Solution

yuliaPel:
[DEMO LINK](https://yuliaPel.github.io/react_redux-list-of-todos/)


src/features/currentTodo.ts
```diff
@@ -1,10 +1,17 @@
-import { createSlice } from '@reduxjs/toolkit';
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
 import { Todo } from '../types/Todo';
 
 const initialState = null as Todo | null;
```

```suggestion
type CurrentTodoState = Todo | null;

const initialState: CurrentTodoState = null;
```

src/features/filter.ts
```diff
@@ -1,12 +1,26 @@
...
-export const filterSlice = createSlice({
+const filterSlice = createSlice({
   name: 'filter',
   initialState,
-  reducers: {},
+  reducers: {
+    setQuery: (state, action: PayloadAction<string>) => ({
+      ...state,
+      query: action.payload,
+    }),
```
dinaminahi, line 16: By the way,  Redux Toolkit's createSlice uses Immer under the hood, which allows you to mutate the state directly without needing to return a new object.
```suggestion
    setQuery: (state, action: PayloadAction<string>) => {
            state.query = action.payload;
    },
```

dinaminahi (approved):
Well done 🥇

PR #976, opened 2024-08-09
add task solution

V-Shut:
[DEMO LINK](https://v-shut.github.io/react_redux-list-of-todos/)


src/App.tsx
```diff
@@ -1,26 +1,95 @@
...
-          <div className="block">
-            <TodoFilter />
-          </div>
+export const App = () => {
+  const [modal, setModal] = useState(false);
+  const [loader, setLoader] = useState(true);
+  const [noFetchedTodos, setNoFetchedTodos] = useState(false);
+  const currentTodo = useAppSelector(store => store.currentTodo);
+  const selectedStatus = useAppSelector(store => store.filter.status);
+  const currentQuery = useAppSelector(store => store.filter.query);
```
dinaminahi, line 22:
```suggestion
  const {status, query} = useAppSelector(store => store.filter);
```

src/features/currentTodo.ts
```diff
@@ -1,10 +1,20 @@
-import { createSlice } from '@reduxjs/toolkit';
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
 import { Todo } from '../types/Todo';
 
 const initialState = null as Todo | null;
```

```suggestion
type CurrentTodoState = Todo | null;

const initialState: CurrentTodoState = null;
```

src/features/filter.ts
```diff
@@ -1,12 +1,25 @@
-import { createSlice } from '@reduxjs/toolkit';
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
 
 const initialState = {
   query: '',
   status: 'all',
```
dinaminahi, line 5: We have enum for this

src/features/filter.ts
```diff
@@ -1,12 +1,25 @@
...
 
-export const filterSlice = createSlice({
+const filterSlice = createSlice({
   name: 'filter',
   initialState,
-  reducers: {},
+  reducers: {
+    setQuery: (state, action: PayloadAction<string>) => ({
+      ...state,
+      query: action.payload,
```
dinaminahi, line 14: By the way,  Redux Toolkit's createSlice uses Immer under the hood, which allows you to mutate the state directly without needing to return a new object.
```suggestion
      state.query = action.payload;
    },
```

dinaminahi (approved):
Good job 🥇 See some advices

PR #980, opened 2024-08-09
react_redux-list-of-todos

LitviVit:
[DEMO LINK](https://LitviVit.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,55 @@
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import { Loader, TodoFilter, TodoList, TodoModal } from './components';
+import { useAppDispatch, useAppSelector } from './app/hooks';
+import { useEffect, useMemo, useState } from 'react';
+import { getTodos } from './api';
+import { todosSlice } from './features/todos';
+import { getPrepearedTodos } from './helpers/getPrepearedTodos';
 
-export const App = () => (
-  <>
-    <div className="section">
-      <div className="container">
-        <div className="box">
-          <h1 className="title">Todos:</h1>
+export const App = () => {
+  const todos = useAppSelector(state => state.todosReducer);
+  const [isLoading, setIsLoading] = useState(false);
+  const filterSelect = useAppSelector(state => state.filterReducer.select);
+  const filterQuery = useAppSelector(state => state.filterReducer.query);
+  const selectedTodo = useAppSelector(state => state.currentTodoReducer);
 
-          <div className="block">
-            <TodoFilter />
-          </div>
+  const dispatch = useAppDispatch();
 
-          <div className="block">
-            <Loader />
-            <TodoList />
+  useEffect(() => {
+    setIsLoading(true);
+
+    getTodos()
+      .then(fetchedTodos => {
+        dispatch(todosSlice.actions.setTodos(fetchedTodos));
+      })
+      .finally(() => setIsLoading(false));
+  }, [dispatch]);
+
+  const filteredTodos = useMemo(() => {
+    return getPrepearedTodos(todos, filterQuery, filterSelect);
+  }, [todos, filterQuery, filterSelect]);
+
+  return (
+    <>
+      <div className="section">
+        <div className="container">
+          <div className="box">
+            <h1 className="title">Todos:</h1>
+
+            <div className="block">
+              <TodoFilter />
+            </div>
+
+            <div className="block">
+              {isLoading && <Loader />}
+
+              <TodoList todos={filteredTodos} />
+            </div>
           </div>
         </div>
       </div>
-    </div>
-
-    <TodoModal />
-  </>
-);
+      {selectedTodo && <TodoModal />}
+    </>
+  );
+};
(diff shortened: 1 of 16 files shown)


src/App.tsx
```diff
@@ -1,26 +1,55 @@
...
-            <TodoFilter />
-          </div>
+  const dispatch = useAppDispatch();
+
+  useEffect(() => {
+    setIsLoading(true);
+
+    getTodos()
+      .then(fetchedTodos => {
+        dispatch(todosSlice.actions.setTodos(fetchedTodos));
```
dinaminahi, line 24: You can export actions from you slice like this

export const { setTodos } = todosSlice.actions;

so here it will be 
```suggestion
        dispatch(setTodos(fetchedTodos));
```

src/app/store.ts
```diff
@@ -1,6 +1,13 @@
...
+import { filterSlice } from '../features/filter';
+import { todosSlice } from '../features/todos';
+import { currentTodoSlice } from '../features/currentTodo';
 
-const rootReducer = combineSlices();
+const rootReducer = combineSlices({
+  todosReducer: todosSlice.reducer,
+  filterReducer: filterSlice.reducer,
+  currentTodoReducer: currentTodoSlice.reducer,
+});
```
dinaminahi, line 10: Better to name 
```suggestion
  todos: todosSlice.reducer,
  filter: filterSlice.reducer,
  currentTodo: currentTodoSlice.reducer,
});
```

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,17 +1,33 @@
 import React from 'react';
+import { FilterOptions } from '../../types';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { filterSlice } from '../../features/filter';
 
 export const TodoFilter: React.FC = () => {
+  const select = useAppSelector(state => state.filterReducer.select);
+  const query = useAppSelector(state => state.filterReducer.query);
```
dinaminahi, line 8:
```suggestion
  const {select, query} = useAppSelector(state => state.filterReducer);
```

src/features/currentTodo.ts
```diff
@@ -1,10 +1,17 @@
-import { createSlice } from '@reduxjs/toolkit';
-import { Todo } from '../types/Todo';
+import { createSlice, PayloadAction, Slice } from '@reduxjs/toolkit';
+import { Todo } from '../types';
 
 const initialState = null as Todo | null;
```

```suggestion
type CurrentTodoState = Todo | null;

const initialState: CurrentTodoState = null;
```

src/features/currentTodo.ts
```diff
@@ -1,10 +1,17 @@
...
 
-export const currentTodoSlice = createSlice({
+export const currentTodoSlice: Slice<Todo | null> = createSlice({
   name: 'currentTodo',
   initialState,
-  reducers: {},
+  reducers: {
+    setCurrentTodo: (_, action: PayloadAction<Todo>) => {
+      return action.payload;
+    },
```
dinaminahi, line 12: 👍

src/features/filter.ts
```diff
@@ -1,12 +1,32 @@
...
+export const filterSlice: Slice<FilterState> = createSlice({
   name: 'filter',
   initialState,
-  reducers: {},
+  reducers: {
+    setQuery(state, action: PayloadAction<string>) {
+      return {
+        ...state,
+        query: action.payload,
+      };
```
dinaminahi, line 22: By the way,  Redux Toolkit's createSlice uses Immer under the hood, which allows you to mutate the state directly without needing to return a new object.
```suggestion
      state.query = action.payload;
    },
```

src/features/filter.ts
```diff
@@ -1,12 +1,32 @@
-import { createSlice } from '@reduxjs/toolkit';
+import { createSlice, PayloadAction, Slice } from '@reduxjs/toolkit';
+import { FilterOptions } from '../types';
 
 const initialState = {
   query: '',
-  status: 'all',
+  select: 'all' as FilterOptions,
```
dinaminahi, line 6:
```suggestion
  select: FilterOptions.All,
```

dinaminahi (changes requested):
Great job 👍 
let's just improve a bit

src/features/filter.ts
LitviVit: image
This is good advice, thanks, but i'm breaking eslint rules this way😣

src/features/currentTodo.ts
LitviVit: Yes, indeed. But I still have to specify it through 'as'. Because I still get type incompatibility

image

LitviVit: Ok, I found the solution. But now I don't specify the type of PayloadAction. The code works, but i think it's not quite right to do it this way 🤔
image

VenomMB: You fix it? because your current version of code is correct

VenomMB (approved):
Great job!

src/features/currentTodo.ts
LitviVit: Yes, I fixed it. But i'm not sure that this code is completely correct. I'd like to see the opinion of the first mentor @dinaminahi

src/features/filter.ts
LitviVit: Yeah, I know. That's why these rules were added to the mate academy's homeworks)
Rules are made to be broken © 🤣👍🏻

PR #983, opened 2024-08-11
add solution task react_redux-list-of-todos

seriuksergii:
[DEMO LINK](https://seriuksergii.github.io/react_redux-list-of-todos/)


src/features/filter.ts
```diff
@@ -1,12 +1,25 @@
...
 };
 
-export const filterSlice = createSlice({
+export const filterSlice: Slice<FilterState> = createSlice({
   name: 'filter',
   initialState,
-  reducers: {},
+  reducers: {
+    setFilter: (state, { payload }: PayloadAction<Status>) => {
+      return { ...state, status: payload };
```
VenomMB, line 19:
```suggestion
      state.query = action.payload;
```
By the way, Redux Toolkit's createSlice uses Immer under the hood, which allows you to mutate the state directly without needing to return a new object.

VenomMB (changes requested):
Great job, let's work on improvements now

PR #992, opened 2024-08-13
add initial solution for the react_redux-list-of-todos

lesdyman:
Hi there! Here is my initial solution for react_redux-list-of-todos. I did what I could with it. You shall find demo link below. Cheers!

[DEMO] (https://lesdyman.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,6 +1,27 @@
 import React from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import { RootState } from '../../app/store';
+import { filterByQuery, filterByStatus } from '../../features/filter';
 
 export const TodoFilter: React.FC = () => {
+  const dispatch = useDispatch();
+  const query = useSelector((state: RootState) => state.filter.query);
+  const status = useSelector((state: RootState) => state.filter.status);
+
+  const handleStatusChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    dispatch(
+      filterByStatus(event.target.value as 'completed' | 'all' | 'active'),
+    );
+  };
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    dispatch(filterByQuery(event.target.value));
+  };
+
+  const handleClearSearch = () => {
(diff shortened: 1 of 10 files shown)


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,14 +1,39 @@
...
+import { filterByQuery, filterByStatus } from '../../features/filter';
 
 export const TodoFilter: React.FC = () => {
+  const dispatch = useDispatch();
+  const query = useSelector((state: RootState) => state.filter.query);
+  const status = useSelector((state: RootState) => state.filter.status);
+
+  const handleStatusChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    dispatch(
+      filterByStatus(event.target.value as 'completed' | 'all' | 'active'),
```
artemforsoff, line 13: you can use an existing type `as Status`

src/components/TodoList/TodoList.tsx
```diff
@@ -1,224 +1,98 @@
...
+                <td className="has-text-right is-vcentered">
+                  <button
+                    data-cy="selectButton"
+                    className="button"
+                    type="button"
+                    onClick={() => dispatch(findCurrentTodo(todo.id))}
+                  >
+                    <span className="icon">
+                      {currentTodo?.id === todo.id ? (
+                        <i className="far fa-eye-slash" />
```
artemforsoff, line 80: better move condition in call `classNames`
like this `classNames('far', currentTodo?.id === todo.id ? 'fa-eye-slash' : 'fa-eye')`

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,40 +1,101 @@
...
+        setLocalLoading(false);
+      }
+    };
+
+    fetchUser();
+  }, [currentTodo, dispatch]);
+
+  const handleCloseModal = () => {
+    dispatch(setCurrentTodo(null));
+    dispatch(clearUser('user cleared'));
```
artemforsoff, line 52:
```
clearUser(state) {
  state.user = null;
  state.loading = false;
  state.error = null;
},
```

action does not accept agruments

artemforsoff (approved):
great work 🔥

PR #993, opened 2024-08-15
solution

getchards:
[DEMO LINK](https://getchards.github.io/react_redux-list-of-todos/)


src/App.tsx
```diff
@@ -1,26 +1,83 @@
...
+            </div>
           </div>
         </div>
       </div>
-    </div>
 
-    <TodoModal />
-  </>
-);
+      {false && <TodoModal />}
```
TarasHoliuk, line 80: You should implement this part as well (and store selected todo)

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -33,6 +62,7 @@ export const TodoFilter: React.FC = () => {
             data-cy="clearSearchButton"
             type="button"
             className="delete"
+            onClick={clearQuery}
```
TarasHoliuk, line 65: this button shouldn't be visible if no text typed in the input - check working version

SanyaBratashchuk (approved):

PR #998, opened 2024-08-20
add solution

AYupko:
[DEMO LINK](https://AYupko.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,224 +1,86 @@
-/* eslint-disable */
 import React from 'react';
+import cn from 'classnames';
+import { useDispatch, useSelector } from 'react-redux';
+import { RootState } from '../../app/store';
+import { filterTodos } from '../../utils/filterTodos';
+import { actions } from '../../features/currentTodo';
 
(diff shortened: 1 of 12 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,224 +1,86 @@
...
-            </td>
+  const todos = useSelector((state: RootState) => state.todos);
+  const { status, query } = useSelector((state: RootState) => state.filter);
+  const currentTodo = useSelector((state: RootState) => state.currentTodo);
+  const dispatch = useDispatch();
 
-            <td className="is-vcentered is-expanded">
-              <p className="has-text-success">quo adipisci enim quam ut ab</p>
-            </td>
+  const visibleTodos = filterTodos(todos, { status, query });
```
alexander-ignatow, line 14: Not critical in this app, but this is the potential case for `useMemo`

alexander-ignatow (approved):
Looks good 👍 
Looking for the next task!

There are some "can do the other way" places in the app
- there is a potential case for `useMemo`
- Potentially filtered todos may be taken to the redux store, but your solution also good

PR #1022, opened 2024-09-09
add solution redux-list-todos

AnnBusya:
[DEMO LINK](https://AnnBusya.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoModal/TodoModal.tsx
+++ b/src/components/TodoModal/TodoModal.tsx
@@ -1,42 +1,74 @@
-import React from 'react';
+import { useEffect, useState } from 'react';
 import { Loader } from '../Loader';
+import { User } from '../../types/User';
+import { useDispatch } from 'react-redux';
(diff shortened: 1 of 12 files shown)


anastasiiavorobiova (changes requested):
Make sure to deploy your solution

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,42 +1,74 @@
...
+        })
+        .finally(() => setIsLoading(false));
+    }
+  }, [currentTodo]);
 
-          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-          <button type="button" className="delete" data-cy="modal-close" />
-        </header>
+  return (
+    !!currentTodo && (
```
etojeDenys, line 31: better to add the condition before the jsx
```suggestion
 if(!currentTodo) {
   return;
 }

  return (
```

etojeDenys (approved):

PR #1036, opened 2024-09-23
add task solution

mykhailonl:
[DEMO](https://mykhailonl.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -16,6 +16,10 @@ export const TodoFilter: React.FC = () => {
     dispatch(setQuery(value.target.value));
   };
 
+  const resetQuery = () => {
+    dispatch(setQuery(''));
+  };
+
   return (
     <form className="field has-addons">
       <p className="control">
@@ -48,7 +52,7 @@ export const TodoFilter: React.FC = () => {
               data-cy="clearSearchButton"
               type="button"
               className="delete"
-              onClick={() => dispatch(setQuery(''))}
+              onClick={resetQuery}
             />
           </span>
         )}
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -69,6 +69,7 @@ export const TodoList: React.FC = () => {
           <tbody>
             {visibleTodos.map(todo => (
               <tr
+                key={todo.id}
                 data-cy="todo"
                 className={classNames('', {
                   'has-background-info-light': todo.id === currentTodo?.id,


src/components/TodoList/TodoList.tsx
```diff
@@ -1,12 +1,55 @@
...
+      const matchesStatus =
+        status === 'all' ||
+        (status === 'active' && !todo.completed) ||
+        (status === 'completed' && todo.completed);
+
+      return matchesQuery && matchesStatus;
+    });
+  };
+
+  const [visibleTodos, setVisibleTodos] = useState(todos);
```
TarasHoliuk, line 30: So todos, query, and status are stored in state (redux state in this case). Taking it into account - the re-render will happen on any of these variables change. It means there is no reason to store visible todos in state - store them in common variable (and useMemo)

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,42 +1,85 @@
...
+              ) : (
+                <strong className="has-text-success">Done</strong>
+              )}
+              {' by '}
+              <a href={`mailto:${currentUser.email}`}>{currentUser.name}</a>
+            </p>
+          </div>
         </div>
-      </div>
+      ) : null}
```
TarasHoliuk, line 82: use `&&` instead of ternary in this case

TarasHoliuk (changes requested):
Good job! A few small comments

TarasHoliuk:
It shouldn't be visible:

image

Check working version

mykhailonl:
> It shouldn't be visible:
> 
> image
> 
> Check working version

cant open an image

src/components/TodoList/TodoList.tsx
```diff
@@ -1,224 +1,121 @@
...
-      </table>
+              </th>
+
+              <th>Title</th>
+              <th> </th>
+            </tr>
+          </thead>
+
+          <tbody>
+            {visibleTodos.map(todo => (
```
anastasiiavorobiova, line 70: Consider using keys while rendering list

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -22,19 +34,24 @@ export const TodoFilter: React.FC = () => {
...
-          />
-        </span>
+        {query && (
+          <span className="icon is-right" style={{ pointerEvents: 'all' }}>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+            <button
+              data-cy="clearSearchButton"
+              type="button"
+              className="delete"
+              onClick={() => dispatch(setQuery(''))}
```
anastasiiavorobiova, line 51:
```suggestion
              onClick={resetQuery}
```

anastasiiavorobiova (changes requested):
Great work! Almost done!

anastasiiavorobiova (approved):
Great work!

PR #1046, opened 2024-09-29
React_Redux_List_Of_Todos

victoriasokolvak:
[DEMO LINK](https://victoriasokolvak.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,47 @@
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import { Loader, TodoFilter, TodoList, TodoModal } from './components';
+import { useDispatch } from 'react-redux';
+import { useAppSelector } from './app/hooks';
+import { useEffect, useState } from 'react';
+import { getTodos } from './api';
+import { setTodos } from './features/todos';
 
-export const App = () => (
-  <>
-    <div className="section">
-      <div className="container">
-        <div className="box">
-          <h1 className="title">Todos:</h1>
+export const App: React.FC = () => {
+  const dispatch = useDispatch();
+  const todos = useAppSelector(state => state.todos);
+  const currentTodo = useAppSelector(state => state.currentTodo);
+  const [loader, setLoader] = useState(false);
 
-          <div className="block">
-            <TodoFilter />
-          </div>
+  useEffect(() => {
+    setLoader(true);
+    getTodos()
+      .then(getTodo => {
+        dispatch(setTodos(getTodo));
+      })
+      .finally(() => setLoader(false));
+  }, [dispatch]);
 
-          <div className="block">
-            <Loader />
-            <TodoList />
+  return (
+    <>
+      <div className="section">
+        <div className="container">
+          <div className="box">
+            <h1 className="title">Todos:</h1>
+
+            <div className="block">
+              <TodoFilter />
+            </div>
+
+            <div className="block">
+              {loader && <Loader />}
+              {!loader && !!todos.length && <TodoList />}
+            </div>
           </div>
         </div>
       </div>
-    </div>
 
-    <TodoModal />
-  </>
-);
+      {currentTodo && <TodoModal />}
+    </>
+  );
+};
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,224 +1,83 @@
 /* eslint-disable */
 import React from 'react';
+import { useDispatch } from 'react-redux';
+import { useAppSelector } from '../../app/hooks';
+import { setCurrentTodo } from '../../features/currentTodo';
+import { Todo } from '../../types/Todo';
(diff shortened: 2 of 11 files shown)


src/App.tsx
```diff
@@ -1,26 +1,47 @@
...
-export const App = () => (
-  <>
-    <div className="section">
-      <div className="container">
-        <div className="box">
-          <h1 className="title">Todos:</h1>
+export const App: React.FC = () => {
+  const dispatch = useDispatch();
+  const todos = useAppSelector(state => state.todos);
+  const currentTodo = useAppSelector(state => state.currentTodo);
```
BudnikOleksii, line 13: It's better to create selectors so you can reuse them
https://redux.js.org/usage/deriving-data-selectors

src/App.tsx
```diff
@@ -1,26 +1,47 @@
...
+  const [loading, isLoading] = useState(false);
```
BudnikOleksii, line 14:
```suggestion
  const [isLoading, setIsLoading] = useState(false);
```
- Boolean variables should start from `is` or `has`
- the second element that returns `useState` is a `setState` function, not a variable

src/components/TodoList/TodoList.tsx
```diff
@@ -1,224 +1,83 @@
...
+            </span>
+          </th>
+          <th>Title</th>
+          <th> </th>
+        </tr>
+      </thead>
+
+      <tbody>
+        {filteredTodos.map(({ id, completed, title, userId }) => (
+          <tr data-cy="todo" key={id}>
```
BudnikOleksii, line 39: Create a todo component

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,42 +1,71 @@
...
+import { useAppSelector } from '../../app/hooks';
+import { setCurrentTodo } from '../../features/currentTodo';
+import { getUser } from '../../api';
+import { User } from '../../types/User';
 
 export const TodoModal: React.FC = () => {
+  const dispatch = useDispatch();
+  const currentTodo = useAppSelector(state => state.currentTodo);
+  const [user, setUser] = useState<User>();
+  const [loading, isLoading] = useState<boolean>(true);
```
BudnikOleksii, line 13: The same regarding naming

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,42 +1,71 @@
...
+
+  const handleClose = () => {
+    dispatch(setCurrentTodo(null));
+  };
+
+  useEffect(() => {
+    isLoading(true);
+    if (currentTodo) {
```
BudnikOleksii, line 21: If there is no `currentTodo` user will see infinite loader

src/utils/filterTodos.ts
```diff
@@ -0,0 +1,22 @@
...
+  query: string,
+  status: Status,
+): Todo[] => {
+  return todos
+    .filter(todo => {
+      if (status === Status.Active) {
+        return !todo.completed;
+      }
+
+      if (status === Status.Completed) {
```
BudnikOleksii, line 15: Use switch case

src/utils/filterTodos.ts
```diff
@@ -0,0 +1,22 @@
+import { Todo } from '../types/Todo';
+import { Status } from '../types/Status';
+
+export const filterTodos = (
+  todos: Todo[],
+  query: string,
+  status: Status,
+): Todo[] => {
+  return todos
+    .filter(todo => {
```
BudnikOleksii, line 10: - if there are no query and status is all you don't need to filter at all
- also, you can try to do it in one filter

BudnikOleksii (changes requested):
Good work 👍 
Left a few suggestions, check them

src/App.tsx
```diff
@@ -1,26 +1,48 @@
...
-          <div className="block">
-            <TodoFilter />
-          </div>
+  useEffect(() => {
+    setIsLoading(true);
+    getTodos()
+      .then(getTodo => {
+        dispatch(setTodos(getTodo));
+      })
+      .finally(() => setIsLoading(false));
```
igorkruz, line 23: It's better to move the logic inside the useEffect hook to a separate function, as it improves code readability, reusability, and makes testing easier. It also helps keep useEffect focused on side effects.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,42 +1,77 @@
...
+
+      return;
+    }
+
+    setIsLoading(true);
+    if (currentTodo) {
+      getUser(currentTodo?.userId)
+        .then(setUser)
+        .finally(() => setIsLoading(false));
+    }
```
igorkruz, line 31: Move the logic inside the `useEffect `hook to a separate function

igorkruz (changes requested):
Good Job 👍 
let's improve your solution

igorkruz (approved):
Well done 🔥

PR #1056, opened 2024-10-03
add task solution

pawelnowicki87:
[DEMO LINK](https://pawelnowicki87.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,6 +1,26 @@
 import React from 'react';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { setQuery, setStatus } from '../../features/filter';
+import { Status } from '../../types/Status';
+
+const statuses: Record<Status, string> = {
+  all: 'All',
+  active: 'Active',
+  completed: 'Completed',
+};
 
 export const TodoFilter: React.FC = () => {
+  const dispatch = useAppDispatch();
+  const query = useAppSelector(state => state.filter.query);
+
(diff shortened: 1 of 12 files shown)


src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,17 +1,39 @@
 import React from 'react';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { setQuery, setStatus } from '../../features/filter';
+import { Status } from '../../types/Status';
+
+const statuses: Record<Status, string> = {
+  all: 'All',
+  active: 'Active',
+  completed: 'Completed',
+};
```
Zibi95, line 10:
```suggestion
const statuses: Record<Status, string> = {
  all: 'All',
  active: 'Active',
  completed: 'Completed',
} as const
```

Zibi95: Assert this kind of objects to `const` so you will get error when you will try to mutate it.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,224 +1,89 @@
...
+                        <i className="far fa-eye" />
+                      )}
+                    </span>
+                  </button>
+                </td>
+              </tr>
+            ))}
+          </tbody>
+        </table>
+      )}
```
Zibi95, line 86: Use destructuring syntax whenever you are able to

Zibi95 (changes requested):
Good job! Left some comments for future reference!

Zibi95 (approved):

PR #1058, opened 2024-10-03
solution

Kozubowicz:
[DEMO LINK](https://Kozubowicz.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoList/TodoList.tsx
+++ b/src/components/TodoList/TodoList.tsx
@@ -1,224 +1,105 @@
 /* eslint-disable */
 import React from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import { RootState } from '../../app/store';
+import { StatusType } from '../../types/Status';
+import { Todo } from '../../types/Todo';
+import { setCurrentTodo } from '../../features/currentTodo';
 
 export const TodoList: React.FC = () => {
+  const todos = useSelector((state: RootState) => state.todos.todos);
+  const status = useSelector((state: RootState) => state.filter.status);
+  const query = useSelector((state: RootState) => state.filter.query);
+
+  const dispatch = useDispatch();
+
+  const filteredTodos = todos
+    .filter(el => {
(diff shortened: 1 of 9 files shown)


src/features/todos.ts
```diff
@@ -1,8 +1,23 @@
...
+
 export const todosSlice = createSlice({
   name: 'todos',
-  initialState: [] as Todo[],
-  reducers: {},
+  initialState,
+  reducers: {
+    setTodos: (state, action: PayloadAction<Todo[]>) => {
+      // eslint-disable-next-line no-param-reassign
+      state.todos = action.payload;
```
Zibi95, line 18: You shouldn't reassign or mutate state in Redux. Always use this syntax to ensure immutability. It's safer, especially if this slice could handle other properties in the future

```
return {
        ...state,
        todos: action.payload,
 };
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,42 +1,72 @@
...
+                )}
+                {' by '}
+                <a href={`mailto:${user.email}`}>{user.name}</a>
+              </p>
+            </div>
+          </div>
+        )
+      )}
     </div>
   );
```
Zibi95, line 71: destructure properties whenever you can

src/components/TodoList/TodoList.tsx
```diff
@@ -1,224 +1,105 @@
...
+                        }
+                      />
+                    </span>
+                  </button>
+                </td>
+              </tr>
+            ))}
+          </tbody>
+        </table>
+      )}
```
Zibi95, line 102: Use destructuring

Zibi95 (changes requested):
Hey. Fix those small things and we are good. While Redux Toolkit allows mutable updates thanks to Immer under the hood, it's still recommended to avoid disabling linting rules. This keeps the codebase consistent and prevents confusion.

PR #1061, opened 2024-10-07
Develop

igorpolianskyi:
[DEMO LINK](https://igorpolianskyi.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,47 @@
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import { Loader, TodoFilter, TodoList, TodoModal } from './components';
+import { useEffect, useState } from 'react';
+import { getTodos } from './api';
+import { useDispatch } from 'react-redux';
+import { useAppSelector } from './app/hooks';
+import { todosSlice } from './features/todos';
 
-export const App = () => (
-  <>
-    <div className="section">
-      <div className="container">
-        <div className="box">
-          <h1 className="title">Todos:</h1>
+export const App = () => {
+  const [listLoading, setListLoading] = useState(false);
+  const dispatch = useDispatch();
 
-          <div className="block">
-            <TodoFilter />
-          </div>
+  const currentTodo = useAppSelector(state => state.currentTodo);
 
-          <div className="block">
-            <Loader />
-            <TodoList />
+  useEffect(() => {
+    setListLoading(true);
+    getTodos()
+      .then(array => {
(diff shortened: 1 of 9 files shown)


TarasHoliuk:
only selected todo should change the icon:

image

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,42 +1,84 @@
...
+    setModalLoading(true);
+
+    if (!currentTodo) {
+      return;
+    }
+
+    getUser(currentTodo.userId)
+      .then(data => setUser({ ...data }))
+      .finally(() => setModalLoading(false));
+  }, []);
```
TarasHoliuk, line 28:
```suggestion
  }, [currenctTodo]);
```
Just add it to the dependencies list)

image

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,14 +1,29 @@
...
+import { filterSlice } from '../../features/filter';
+import { useAppSelector } from '../../app/hooks';
 
 export const TodoFilter: React.FC = () => {
+  const { query, status } = useAppSelector(state => state.filter);
+
+  const dispatch = useDispatch();
+
+  const handleStatus = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    dispatch(filterSlice.actions.setStatus(event.target.value));
```
TarasHoliuk, line 12: It's better to export actions from the slice file and import them here, in slice:

```
export const filterActions = filterSlice.actions;
```
or even 
```
export const { setStatus, setQuery } = filterSlice.actions;
```

src/App.tsx
```diff
@@ -1,26 +1,47 @@
...
+  const currentTodo = useAppSelector(state => state.currentTodo);
+
+  useEffect(() => {
+    setListLoading(true);
+    getTodos()
+      .then(array => {
+        dispatch(todosSlice.actions.setTodos(array));
+      })
+      .finally(() => setListLoading(false));
+  }, []);
```
TarasHoliuk, line 23: to avoid lint error it's better just add `dispatch` to dependencies list (it's memorized too, so actually it will work the same), or at leat disable lint for this row

TarasHoliuk (changes requested):
GJ! Added a few small comments

etojeDenys (changes requested):
image
still not fixed

2. you don't need to show the TodoList on loading

https://github.com/user-attachments/assets/cc7d9f71-7676-4531-aa52-c5ee14443be3

SanyaBratashchuk (approved):

PR #1117, opened 2024-12-07
add solution

modeltoIT:
[DEMO LINK](https://modeltoit.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/components/TodoFilter/TodoFilter.tsx
+++ b/src/components/TodoFilter/TodoFilter.tsx
@@ -1,6 +1,33 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { FilterBy } from '../../types/FilterBy';
+import { useAppDispatch } from '../../hooks/useAppDispatch';
+import { filterSlice } from '../../features/filter';
+import { useAppSelector } from '../../hooks/useAppSelector';
+import { useDebounce } from '../../hooks/useDebounce';
 
 export const TodoFilter: React.FC = () => {
+  const [query, setQuery] = useState('');
+  const { status } = useAppSelector(state => state.filter);
+  const dispatch = useAppDispatch();
+  const debouncedValue = useDebounce(query, 500);
+
+  const selectValue = (value: FilterBy) => {
+    dispatch(filterSlice.actions.filterStatus(value));
+  };
+
+  const search = (value: string) => {
+    setQuery(value);
+  };
+
+  const resetQuery = () => {
+    setQuery('');
+    dispatch(filterSlice.actions.filterQuery(''));
+  };
+
+  useEffect(() => {
+    dispatch(filterSlice.actions.filterQuery(debouncedValue.toLowerCase()));
+  }, [debouncedValue]);
+
   return (
     <form
       className="field has-addons"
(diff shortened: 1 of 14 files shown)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,224 +1,61 @@
...
-            </td>
-          </tr>
+  const todos = useAppSelector(state => state.todos);
+  const { status, query } = useAppSelector(state => state.filter);
 
-          <tr data-cy="todo" className="has-background-info-light">
-            <td className="is-vcentered">3</td>
-            <td className="is-vcentered"> </td>
+  const filteredTodos = todos.filter(todo => {
+    let passedQuery = false;
```
TarasHoliuk, line 12: it can be const variable instead of let

src/features/filter.ts
```diff
@@ -1,12 +1,20 @@
...
-  status: 'all',
+  status: FilterBy.All,
 };
 
 export const filterSlice = createSlice({
   name: 'filter',
   initialState,
-  reducers: {},
+  reducers: {
+    filterStatus(state, { payload }: PayloadAction<FilterBy>) {
```
TarasHoliuk, line 13:
```suggestion
    setStatus(state, { payload }: PayloadAction<FilterBy>) {
```
this action sets the filter status, so should be named accordingly
Same below

src/features/currentTodo.ts
```diff
@@ -1,10 +1,17 @@
...
 import { Todo } from '../types/Todo';
 
 const initialState = null as Todo | null;
 
 export const currentTodoSlice = createSlice({
   name: 'currentTodo',
   initialState,
-  reducers: {},
+  reducers: {
+    open: (_, { payload }: PayloadAction<Todo>) => {
```
TarasHoliuk, line 10: in context of this slice `open` is not correct name for the action - it sets the current todo
So it should be `setCurrentTodo`

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,42 +1,71 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { Loader } from '../Loader';
+import { useAppSelector } from '../../hooks/useAppSelector';
+import { getUser } from '../../api';
+import { User } from '../../types/User';
+import { useAppDispatch } from '../../hooks/useAppDispatch';
+import { currentTodoSlice } from '../../features/currentTodo';
```
TarasHoliuk, line 7: you can export all actions (using named export) separately from the slice file and import them here

TarasHoliuk (changes requested):

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,17 +1,50 @@
...
+  const [query, setQuery] = useState('');
+  const { status } = useAppSelector(state => state.filter);
+  const dispatch = useAppDispatch();
+  const debouncedValue = useDebounce(query, 500);
+
+  const selectValue = (value: FilterBy) => {
+    dispatch(filterSlice.actions.filterStatus(value));
+  };
+
+  const search = (value: string) => {
```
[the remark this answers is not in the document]
modeltoIT: I use the local state there to implement useDebounce

anastasiiavorobiova (approved):

PR #1168, opened 2025-01-29
solution

Daryna-Kukharets:
[DEMO LINK](https://Daryna-Kukharets.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/app/functions.ts
+++ b/src/app/functions.ts
@@ -1,0 +1,29 @@
+import { Todo } from '../types/Todo';
+
+export const filterTodos = (
+  todos: Todo[],
+  status: string,
+  query: string,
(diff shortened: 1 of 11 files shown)


src/app/functions.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const filterTodos = (
+  todos: Todo[],
+  status: string,
+  query: string,
+): Todo[] => {
+  let filteredTodos;
+
+  switch (status) {
+    case 'active':
```
maxim2310, line 11: would be better to use enum for status

maxim2310 (approved):
Looks good for me, just try to use enum for such stuff as i commented

PR #1203, opened 2025-03-01
Task solution

Dominik-Dudek-code:
- [DEMO LINK](https://Dominik-Dudek-code.github.io/react_redux-list-of-todos/)


src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,42 +1,82 @@
...
+  const dispatch = useAppDispatch();
+  const { todo: currentTodo, selectedTodoUser } = useAppSelector(
+    state => state.currentTodo,
+  );
+
+  useEffect(() => {
+    setIsModalLoading(true);
+    if (!selectedTodoUser) {
+      return;
+    }
```
Zibi95, line 21: it probably won't happen but it looks like potential issue. You set `isLoading` to true and then return early

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,42 +1,82 @@
...
 
-      <div className="modal-card">
-        <header className="modal-card-head">
-          <div
-            className="modal-card-title has-text-weight-medium"
-            data-cy="modal-header"
-          >
-            Todo #3
-          </div>
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
Zibi95, line 53: You can fix by adding aria-label to the button

Zibi95 (changes requested):

PR #1213, opened 2025-03-14
Develop

aMtiIV:
[DEMO LINK](https://aMtiIV.github.io/react_redux-list-of-todos/)


aMtiIV:
src/components/Loader/Loader.scss is needed for loading animation. This file was provided in the forked repository. It's not my file so I think it should stay in the project.

natalia-klonowska (changes requested):
notification "There are no todos matching current filter criteria" should only appear after loader for todoList has finished to avoid flickering before the content is fully loaded

danon321 (approved):

PR #1368, opened 2025-09-10
Develop

Danylo-Kachaniuk:
[DEMO LINK](https://Danylo-Kachaniuk.github.io/react_redux-list-of-todos/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,44 @@
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import { Loader, TodoFilter, TodoList, TodoModal } from './components';
+import { useEffect } from 'react';
+import { useAppDispatch, useAppSelector } from './app/hooks';
+import { getTodos } from './api';
+import { actions as todosActions } from './features/todos';
 
-export const App = () => (
-  <>
-    <div className="section">
-      <div className="container">
-        <div className="box">
-          <h1 className="title">Todos:</h1>
+export const App = () => {
+  const user = useAppSelector(state => state.currentTodoSlice.assignedUser);
+  const loader = useAppSelector(state => state.todos.loader);
 
-          <div className="block">
-            <TodoFilter />
-          </div>
+  const dispatch = useAppDispatch();
 
-          <div className="block">
-            <Loader />
-            <TodoList />
+  useEffect(() => {
+    dispatch(todosActions.setIsTodoLoading(true));
+
+    getTodos()
+      .then(todosFromServer => {
+        dispatch(todosActions.setTodos(todosFromServer));
+      })
+      .finally(() => dispatch(todosActions.setIsTodoLoading(false)));
+  }, [dispatch]);
+
+  return (
+    <>
+      <div className="section">
+        <div className="container">
+          <div className="box">
+            <h1 className="title">Todos:</h1>
+
+            <div className="block">
+              <TodoFilter />
+            </div>
+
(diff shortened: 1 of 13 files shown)


src/App.tsx
```diff
@@ -1,26 +1,44 @@
...
+  const todo = useAppSelector(state => state.currentTodoSlice.currentTodo);
+  const loader = useAppSelector(state => state.todos.loader);
 
-          <div className="block">
-            <TodoFilter />
-          </div>
+  const dispatch = useAppDispatch();
+
+  useEffect(() => {
+    dispatch(todosActions.setIsTodoLoading(true));
```
FaiHamid, line 16: The value of the loader is only used in this component, so it would be more logical to do this via useState.

src/components/TodoList/TodoList.tsx
```diff
@@ -1,224 +1,54 @@
...
-
-            <td className="has-text-right is-vcentered">
-              <button data-cy="selectButton" className="button" type="button">
-                <span className="icon">
-                  <i className="far fa-eye" />
-                </span>
-              </button>
-            </td>
-          </tr>
+  const filteredTodos = todos.filter(todo => {
```
FaiHamid, line 13: Here, it would be better to move this logic into a separate function.
For example: 
`const filteredTodos  = getFilteredTodos(todos, status, query);`
Then you simply transfer this logic to the getFilteredTodos function.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,42 +1,69 @@
...
-            data-cy="modal-header"
-          >
-            Todo #3
-          </div>
-
-          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-          <button type="button" className="delete" data-cy="modal-close" />
-        </header>
+      {loader ? (
+        <Loader />
```
FaiHamid, line 28: For some reason, this loader is not visible at all on the demo link, i.e., the window appears immediately and then the user is added to it.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,42 +1,69 @@
...
+  const user = useAppSelector(state => state.currentTodoSlice.assignedUser);
+  const dispatch = useAppDispatch();
+
+  useEffect(() => {
+    if (todo) {
+      getUser(todo.userId)
+        .then(userFromServer => {
+          dispatch(currentTodoActions.setAssignedUser(userFromServer));
+        })
+        .finally(() => dispatch(currentTodoActions.setUserIsLoading(false)));
```
FaiHamid, line 19: The same applies here; the value for the loader can simply be added to useState.

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,42 +1,69 @@
...
+import React, { useEffect } from 'react';
 import { Loader } from '../Loader';
+import { useAppDispatch, useAppSelector } from '../../app/hooks';
+import { actions as currentTodoActions } from '../../features/currentTodo';
+import { getUser } from '../../api';
 
 export const TodoModal: React.FC = () => {
+  const loader = useAppSelector(state => state.currentTodoSlice.isUserLoading);
+  const todo = useAppSelector(state => state.currentTodoSlice.currentTodo);
+  const user = useAppSelector(state => state.currentTodoSlice.assignedUser);
```
FaiHamid, line 10: Here also you can use for user useState.

FaiHamid (changes requested):
Greate job!!
Let`s just fix few details:

PR #1369, opened 2025-09-11
implement features with redux

YuliaProkipchuk:
[DEMO LINK](https://YuliaProkipchuk.github.io/react_redux-list-of-todos/)


YuliaProkipchuk:
@mateacademy-ai-mentor i fixed this according to your recommendations but useDispatch.withTypes and useSelector.withTypes are used directly from Redux, so i don't think that would be an issue.

PR #1380, opened 2025-09-22
Develop

creepycryptocock:
https://creepycryptocock.github.io/react_redux-list-of-todos/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,19 +1,27 @@
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import { Loader, TodoFilter, TodoList, TodoModal } from './components';
-import React, { useEffect } from 'react';
+import React, { useEffect, useState } from 'react';
 import { getTodos } from './api';
-import todosSlice from './features/todos';
+import { todosSlice } from './features/todos';
 import { useAppDispatch } from './hooks';
 
 export const App = () => {
   const dispatch = useAppDispatch();
+  const [isLoading, setIsLoading] = useState(false);
 
   useEffect(() => {
     async function fetchTodos() {
-      const todos = await getTodos();
+      try {
+        setIsLoading(true);
+        const todos = await getTodos();
 
-      dispatch(todosSlice.actions.setTodos(todos));
+        dispatch(todosSlice.actions.setTodos(todos));
+      } catch {
+        setIsLoading(false);
+      } finally {
+        setIsLoading(false);
+      }
     }
 
     fetchTodos();
@@ -31,7 +39,7 @@ export const App = () => {
             </div>
 
             <div className="block">
-              <Loader />
+              {isLoading && <Loader />}
               <TodoList />
             </div>
           </div>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -18,14 +18,14 @@ export const App = () => {
 
         dispatch(todosSlice.actions.setTodos(todos));
       } catch {
-        setIsLoading(false);
+        return 'Something went wrong';
       } finally {
         setIsLoading(false);
       }
     }
 
     fetchTodos();
-  }, []);
+  }, [dispatch]);
 
   return (
     <>
@@ -40,7 +40,7 @@ export const App = () => {
 
             <div className="block">
               {isLoading && <Loader />}
-              <TodoList />
+              { !isLoading && <TodoList />}
             </div>
           </div>
         </div>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -40,7 +40,7 @@ export const App = () => {
 
             <div className="block">
               {isLoading && <Loader />}
-              { !isLoading && <TodoList />}
+              <TodoList isLoading={isLoading} />
             </div>
           </div>
         </div>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,44 @@
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import { Loader, TodoFilter, TodoList, TodoModal } from './components';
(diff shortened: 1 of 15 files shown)


src/App.tsx
```diff
@@ -1,26 +1,52 @@
...
-          </div>
+  useEffect(() => {
+    async function fetchTodos() {
+      try {
+        setIsLoading(true);
+        const todos = await getTodos();
+
+        dispatch(todosSlice.actions.setTodos(todos));
+      } catch {
+        setIsLoading(false);
```
TarasHoliuk: You don't need it here, because you're already doing it in the finally block

src/App.tsx
```diff
@@ -1,26 +1,52 @@
...
+      } finally {
+        setIsLoading(false);
+      }
+    }
+
+    fetchTodos();
+  }, []);
```
TarasHoliuk: Знімок екрана 2025-09-22 о 19 21 10

Just add it) It won't change the behavior (because dispatch function is memoized and doesn't change), just to fix lint warning

src/App.tsx
```diff
@@ -1,26 +1,52 @@
...
+            <div className="block">
+              <TodoFilter />
+            </div>
 
-          <div className="block">
-            <Loader />
-            <TodoList />
+            <div className="block">
+              {isLoading && <Loader />}
+              <TodoList />
```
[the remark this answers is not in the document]
TarasHoliuk: Good comment - now, during loading, you also show "No todos" message. It shouldn't work like this

Знімок екрана 2025-09-22 о 19 19 16

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -1,14 +1,24 @@
 import React from 'react';
+import { useAppDispatch, useAppSelector } from '../../hooks';
+import { setQuery, setStatus, clearFilter } from '../../features/filter';
 
 export const TodoFilter: React.FC = () => {
+  const dispatch = useAppDispatch();
+  const status = useAppSelector(state => state.filter.status);
+  const query = useAppSelector(state => state.filter.query);
```
TarasHoliuk:
```suggestion
  const { status, query } = useAppSelector(state => state.filter);
```
nit: You can get 2 properties with one selector

src/components/TodoFilter/TodoFilter.tsx
```diff
@@ -22,18 +32,29 @@ export const TodoFilter: React.FC = () => {
...
+              onClick={() => dispatch(clearFilter())}
+            />
+          )}
+          {/* <button
             data-cy="clearSearchButton"
             type="button"
             className="delete"
-          />
+            onClick={() => dispatch(clearFilter())}
+          /> */}
```
TarasHoliuk: Remove such comments

src/components/TodoList/TodoList.tsx
```diff
@@ -1,14 +1,46 @@
...
+  const dispatch = useAppDispatch();
+  const currentTodo = useAppSelector(state => state.currentTodo);
+
+  const visibleTodos = todos.filter(todo => {
+    let matchQuery = query ? todo.title.toLowerCase().includes(query.toLowerCase()) : true;
+    let matchFilter;
+
+    // if (query) {
+    //   matchQuery = todo.title.includes(query);
+    // }
```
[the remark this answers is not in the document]
TarasHoliuk: Agree

src/components/TodoList/TodoList.tsx
```diff
@@ -25,200 +57,31 @@ export const TodoList: React.FC = () => {
...
-                  <i className="far fa-eye" />
+                  <i className={`${currentTodo?.id === todo.id ? "far fa-eye-slash" : "far fa-eye" }`}/>
                 </span>
               </button>
             </td>
           </tr>
+          ))}
         </tbody>
-      </table>
+      </table>:  <p className="notification is-warning">
```
TarasHoliuk: fix formatting here. Example with ternary and multi-lines results:

```
{condition ? (
  <div>
    Example
  </div>
) : (
  <div>
    Example
  </div>
)}
```

src/components/TodoModal/TodoModal.tsx
```diff
@@ -1,39 +1,83 @@
...
+      {isLoading && <Loader />}
 
       <div className="modal-card">
         <header className="modal-card-head">
           <div
             className="modal-card-title has-text-weight-medium"
             data-cy="modal-header"
           >
-            Todo #3
+            {`Todo #${currentTodo?.id}`}
```
TarasHoliuk:
```suggestion
            {`Todo #${currentTodo.id}`}
```
You've already checked `currentTodo` above. Same everywhere below

TarasHoliuk (changes requested):
Good job! Left a few comments to improve your code :)

etojeDenys (changes requested):
almost done

fix the problem with this message. it should not be displayed if the loader is shown
image

creepycryptocock:
I don't see this message when loading the page. I fixed it with {!isLoading && <TodoList />}, but I suppose it is not correct way

src/components/TodoList/TodoList.tsx
```diff
@@ -25,200 +54,34 @@ export const TodoList: React.FC = () => {
...
             </td>
           </tr>
+          ))}
         </tbody>
       </table>
+      ) : (
+        <p className="notification is-warning">
+        There are no todos matching current filter criteria
+      </p>
+    )}
```
etojeDenys, line 84: try to add here !isLoad &&

etojeDenys (changes requested):
problem is still not fixed

vitalii-bondarenko-ua (approved):
Looks good.

PR #1408, opened 2025-11-23
add redux

Blervin1:
[DEMO LINK](https://blervin1.github.io/react_redux-list-of-todos/)


src/components/TodoList/TodoList.tsx
```diff
@@ -1,224 +1,104 @@
...
+                    type="button"
+                    onClick={() => {
+                      if (openTodoId === todo.id) {
+                        setSelectedTodo(null);
+                        setOpenTodoId(null);
+                      } else {
+                        setSelectedTodo(todo);
+                        setOpenTodoId(todo.id);
+                      }
+                    }}
```
FaiHamid, line 84: If the logic takes up more than one line, it is better to move it to a separate function.

FaiHamid (changes requested):
Well done 👏
Just need to fix one little thing.

